PR #2, opened 2025-10-14
Update some snapshots and tests so they pass again

david-pl:


david-pl:
@Roger-luo the last commit fixes a bug in the Hadamard gate.

PR #3, opened 2025-10-24
Draft impl for ::PauliSum.trace(::PauliSum)

david-pl:
Probably shouldn't merged, but just so we can discuss how to best do this. The problem is that the generic trait implementation for `PauliSum` also covers the case `PauliSum.trace(PauliSum)`, but the typing is off.

This also requires some changes to the underlying `trace` methods on `D/HashMap`, so this might affect performance.


david-pl:
There is now a specific `PauliSum::overlap(other: PauliSum)` method that was added in #20 .

PR #18, opened 2025-11-17
Fix default capacity

david-pl:
* Sets the capacity in trotter example & benchmarks
* Adds a benchmark for trotter as function of qubit number
* Changes the default capacity for strategies

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /target
+.DS_Store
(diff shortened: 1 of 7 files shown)


Roger-luo (approved):
I think I set the capacity because you don’t always initialize with `n` terms. 

The always-work solution is still running a compiler path (a.k.a propagating a number to count the maximum branches) which we need to implement later for surrogates anyways. 

But in terms of the runtime a good enough heurestic and some docstring is sufficient - so I think this PR fixes the problem.

Issue #19, opened 2025-11-28
Python bindings for ppvm-runtime


david-pl:
In order to make this simulator available to users, we need to expose Python bindings. The plan here is to expose things directly via pyo3. For a first version, this would make it immediately possible to implement an interpreter in bloqade-circuit (similar to what we have for PyQrack).

While this will result in some runtime overhead, that should be negligible for shallow circuits. Also, it's the fastest way to make this available to users.
Should the runtime become an issue, we can introduce a circuit object (or, really, the byte code from pao) so all work is done in rust.

david-pl:
So, I'm looking into this and I'm hitting a fundamental issue of rust's compile time guarantees vs. Python's "let's just run it and see what happens" attitude. In particular, we have statically sized hashes, e.g. `ByteF64<const N: usize, ...>`.

The problem becomes apparent when considering that `pyo3` does not support generic types, see here https://pyo3.rs/v0.27.1/class.html#no-generic-parameters.

If we want to expose `PauliSum<T: Config>`, we need to create a wrapper `struct` for all values of `T` that we want to expose. This includes:

* `Map` types, which in turn contain the hash type.
* `Coeff` types, though we can just fix that to `f64`.
* `Strategy` types.

**Side note**: if we are creating a wrapper struct around `PauliSum` anyway, we might as well just think about immediately going for a `Circuit` object. However, the fundamental issue with the hash size will remain in any case.

Now, the fundamental problem is that the size of the hash depends on the number of qubits, which we won't know from the python side until runtime. The only option I see here is to generate a fixed number of wrappers for different sizes. Then, we can implement another thin wrapper in Python that picks the correct interface based on the provided number of qubits.

We can use macros for generating the interfaces and even take it to the extreme and generate an interface for _every_ possible value of `N`. However, I'm not sure about the runtime overhead vs. the compile time when generating thousands of structs. We have to look into that.

A draft is here: #20 .

cc. @Roger-luo

PR #20, opened 2025-11-28
Add python bindings

david-pl:
Draft implementation of #19, see the issue for more details on the basic problems.

Example usage:

```shell
# cd ppvm-py
uv sync
source .venv/bin/activate
maturin develop --uv
python test_basics.py
```

So far, only a single map type (indexmap) is supported, but the macro would make it easy to add more. The more fundamental issue of compile time size for hashes needs some discussion, I think. For now, I went for the approach where I just generate very few interfaces at the cost of having a potentially large overhead when picking the closest one.

Also, to make my life easier I only support one strategy, `CombinedStrategy<CoefficientThreshold, MaxPauliWeight>`. While I think we should always have `CoefficientThreshold` in there (at least with `min_abs_coeff = 0.0`), there will be some overhead when including a `MaxPauliWeight` with a limit that is so large it never truncates anything, which is essentially how I disable it at the moment.


crates/ppvm-py/src/interface.rs
```diff
@@ -0,0 +1,155 @@
...
+                self.inner.truncate();
+            }
+        }
+
+    };
+}
+
+// TODO: use macro to loop over these? Let's check performance first
+type IndexMapFxHash10 =
+    config::indexmap::ByteFxHashF64<10, CombinedStrategy<CoefficientThreshold, MaxPauliWeight>>;
```
Roger-luo: I think a macro loop might be better, we should also provide some 2^n bindings.

Roger-luo (approved):
I think the approach is what I envisioned, we can add a bit more bindings accordingly - we cannot provide arbitrary byte size because every different size will trigger compilation to be efficient (unless we provide a separate `PauliWord` that allows the bitstring to be dynamic)

bikeshedding comment - can we rename `ppvm-py` to `ppvm-python`?

crates/ppvm-py/src/interface.rs
david-pl: As we discussed some time ago: a macro loop here is a bit tricky, since we'd need the `paste` crate in order to generate the names for the different `IndexMapFxHashXX` interfaces. However, due to the order in which macros are expanded, `pyo3` doesn't like e.g. `#[pyclass]` annotations inside a `paste!` macro.

So we'd have to write our own `paste!` macro in order to do this. For now, let's stick with having the interfaces we need hard-coded.

david-pl:
@Roger-luo I think this should be ready to go. To add some details there's now two python packages:

* `ppvm-python-native`: This is just the pyo3 generated package which provides the interfaces.
* `ppvm-python`: This is an actual python package, which adds `ppvm-python-native` as a dependency. I wrapped all the interfaces in a single `PauliSum` class there.

AFAIK, there's not way to bundle a small python wrapper with the pyo3 generated one directly, so this is the way to go here. It's not quite ideal, since we now have three places where we define all the gates, which we need to keep in sync:
* The actual `ppvm` implementation.
* The pyo3 interface (one macro).
* The `PauliSum` wrapper.

However, I'd argue that once we have things stable enough, adding new gates etc. will happen rarely.

Note, that you can already install the package using pip (need to update the URL once it's merged):

```
uv pip install git+https://github.com/QuEraComputing/ppvm.git@david/python-bindings#subdirectory=ppvm-python
```

Then you can do something like

```python
from ppvm_python import PauliSum

state = PauliSum.from_str("ZZ")

# NOTE: the above is equivalent to
# state = PauliSum(n_qubits=2, terms=["ZZ"], min_abs_coeff=1e-10, max_pauli_weight=None)

state.cnot(0, 1)
state.h(0)

print(state)
print(state.overlap_with_zero())
```

We can use this to have people test it internally.

There's also a draft of an interpreter, so we can run this directly on squin kernels here: https://github.com/QuEraComputing/bloqade-internal/pull/162

ppvm-python/pyproject.toml
```diff
@@ -0,0 +1,28 @@
+[project]
+name = "ppvm-python"
```
Roger-luo: let's just call it ppvm

PR #25, opened 2026-01-12
naive tableau propagation

Roger-luo:
I coded this with codex, I'm not 100% my reviews of the implementation covers everything. I tried to generate more tests to make sure things work. Would be nice to have a second look.

I haven't implemented the sampling and canonicalization, but I think these two should be relatively simple to implement later.


crates/ppvm-tableau/src/tableau.rs
```diff
@@ -0,0 +1,166 @@
...
+    xor_phase_with, xor_phase_with_scalar, xor_phase_with_xor, xor_phase_with_xor_scalar,
+};
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Tableau {
+    pub(crate) n_qubits: usize,
+    pub(crate) n_words: usize,
+    pub(crate) x: Vec<u64>,
+    pub(crate) z: Vec<u64>,
+    pub(crate) phase: Vec<u64>,
```
Roger-luo, line 13: this is pretty much what were missing from STIM to allow us doing the propagation.

crates/ppvm-tableau/src/tableau.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use crate::simd::{
+    xor_phase_with, xor_phase_with_scalar, xor_phase_with_xor, xor_phase_with_xor_scalar,
+};
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Tableau {
+    pub(crate) n_qubits: usize,
+    pub(crate) n_words: usize,
+    pub(crate) x: Vec<u64>,
```
rafaelha, line 11: Essentially, this stores stabilizers as Pauli strings using [x|z|phase]

Image

However, this is only one half of the tableau. The tableau consists of both stabilizers and destabilizers. See https://arxiv.org/pdf/quant-ph/0406196

Destabilizers are useful because they reduce the complexity for computing measurements from O(n^3) to O(n^2)

To add destabilizers: just add n more rows and initialize the x part to identity and the z part to 0. All Clifford operations stay as they are (but columns are now twice as long)

crates/ppvm-tableau/src/tableau.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn cz(&mut self, _control: usize, _target: usize) {
+        unimplemented!("CZ not yet implemented for Tableau");
+    }
+}
+
+fn words_for(n_qubits: usize) -> usize {
+    (n_qubits + 63) / 64
+}
+
+fn set_column_bit(words: &mut [u64], n_words: usize, col: usize, row: usize, value: bool) {
```
rafaelha, line 157: The matrices are stored in column-major order, and bitpacked across columns.

This is optimal for Clifford operations. However, it's not ideal for non-deterministic measurements, where one needs row operations.

Is that a design decision that we want to make?

STIM actually transposes between column and row major order. See section 5.1 of https://arxiv.org/pdf/2103.02202

crates/ppvm-tableau/src/simd.rs
```diff
@@ -0,0 +1,354 @@
+#[cfg(all(target_arch = "x86_64", target_feature = "avx2"))]
+#[inline(always)]
+pub(crate) fn xor_phase_with(phase: &mut [u64], bits: &[u64]) {
```
rafaelha, line 3: All of these functions should have unit tests. But CI would need to support compilation for all these targets

crates/ppvm-tableau/src/tableau.rs
```diff
@@ -0,0 +1,166 @@
...
+        self.n_qubits
+    }
+
+    pub fn x(&mut self, qubit: usize) {
+        let base = qubit * self.n_words;
+        let z = &self.z[base..base + self.n_words];
+        xor_phase_with(&mut self.phase, z);
+    }
+
+    pub fn x_scalar(&mut self, qubit: usize) {
```
rafaelha, line 44: What's the reason for the `_scalar` gates to be around?

The scalar case will be used as a fallback when no SIMD instructions are available, so can these explicit scalar instructions just be removed?

crates/ppvm-tableau/src/tableau.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn h(&mut self, index: usize) {
+        Tableau::h(self, index);
+    }
+
+    fn s(&mut self, index: usize) {
+        Tableau::s(self, index);
+    }
+
+    fn cnot(&mut self, _control: usize, _target: usize) {
+        unimplemented!("CNOT not yet implemented for Tableau");
```
rafaelha, line 145: `cnot` should be implemented as well using: 

Image (from https://arxiv.org/pdf/quant-ph/0406196)

crates/ppvm-tableau/src/tableau.rs
```diff
@@ -0,0 +1,166 @@
...
+        xor_phase_with(&mut self.phase, x);
+    }
+
+    pub fn z_scalar(&mut self, qubit: usize) {
+        let base = qubit * self.n_words;
+        let x = &self.x[base..base + self.n_words];
+        xor_phase_with_scalar(&mut self.phase, x);
+    }
+
+    pub fn h(&mut self, qubit: usize) {
```
rafaelha, line 76: `h` and `s` don't have SIMS versions.

Here, `h` and `h_scalar` implementations are identical.

rafaelha (changes requested):

crates/ppvm-tableau/src/simd.rs
david-pl: I agree. Especially given the use of `unsafe` here.

Maybe we should consider putting them in a separate crate and set up CI for that crate only to run on all these architectures?

Actually, are these functions for performance optimizations only or is there real functionality added? If it's just optimizations, do they really need to be part of the first version?

crates/ppvm-tableau/src/tableau.rs
david-pl: Also, `cz` to support the native gate set.

david-pl (commented):
I think generally speaking this is a solid starting point. One thing that's also missing still is some convenient way to compute the overlap, e.g. a `trace` impl.

PR #26, opened 2026-01-15
Add simple qubit loss channel

david-pl:
Adds a "naive" qubit loss channel. The channel reduces the trace, which is equivalent to losing probability out of the qubit subspace.

Specifically, it is equivalent to counting every shot in a state vector simulator as contributing 0 to an expectation value. For example

```python
@squin.kernel
def ghz():
    q = squin.qalloc(2)

    squin.h(q[0])
    pl1 = 0.3
    squin.qubit_loss(pl1, q[0])

    squin.cx(q[0], q[1])
    pl2 = 0.2
    squin.qubit_loss(pl2, q[0])
    squin.qubit_loss(pl2, q[1])

    return q


sim = StackMemorySimulator(min_qubits=2, loss_m_result=Measurement.Zero)
nshots = 2000

Z = np.array([[1, 0], [0, -1]])
ZZ = np.kron(Z, Z)

avg = 0.0
for _ in range(nshots):
    res = sim.run(ghz)

    if any(not q.is_active() for q in res):
        # NOTE: count shots with loss as zero
        continue

    ket = res[0].sim_reg.out_ket()
    exp = np.vdot(ket, ZZ @ ket)
    avg += exp.real / nshots

print(avg)
```

While this somewhat makes sense, it's not quite what we need to emulate what happens on hardware. The problem is that we never really measure an expectation value, but really just obtain a bunch of bit strings from which we reconstruct it. A lost qubit is counted as 0 rather than neglecting the contribution entirely. Unfortunately, I didn't manage to implement this (yet).

The above is correct during gate application in that it accurately captures that lost qubits no longer participate in subsequent gates. However, we'd need some way to then at the end of the circuit fold all lost population into the 0 state. I'm not quite sure what this channel has to look like, but I am pretty sure that it branches causing exponential growth, since we have to bias Z towards the 0 state, which means you have something akin to amplitude damping

$Z \Rightarrow (1 - p_L) Z + p_L \mathbb{1}$

Another source of confusion here is what "at the end" actually means. Since we are propagating backwards through the circuit, we'd need to apply this "reset to 0" channel at the very beginning. However, the probability has to be the accumulated loss at the end of the circuit (when going forward), I think, since we need to fold all population that was lost back into 0. This can be obtained by using forward data flow analysis (ignoring control flow for the time being).

Alternatively, I think it might be possible to only reset the population (conserving the trace) when taking the overlap via `state.trace(&zero_pattern)`. However, there are again exponentially many terms then, I think.

Maybe this can be implemented efficiently using tableaux? Not sure. cc @Roger-luo


david-pl:
Superseded by #28

Issue #30, opened 2026-01-23
Support for correlated loss


david-pl:
Building on #29 (and #28 in particular), we can also look into the support for correlated loss. This will again require some research, since it's not quite clear what the underlying process looks like and how the mappings of the Pauli pairs is defined.

david-pl:
Just to add some details, I did some research and it seems the Pauli channel for correlated loss is

$$\mathcal{E}(P) = (1 - p_L) P , ~ P \neq LL$$
$$\mathcal{E}(LL) = p_L \left( II + IL + LI\right) + LL$$

Here $I = \text{diag}(1, 1, 0)$ is the qubit subspace density matrix and $L$ is the projector on the leakage state. I.e., we use the same definitions and basis as in #28. So, we again have some branching, but all branches are proportional to $p_L$, meaning they should be easy to truncate.

Physically, this process corresponds to decay of the qubit states to a leakage state with the jump operators

$$ J_{ij} = \sqrt{\gamma_L} |LL\rangle\langle ij|,$$

where $|ij\rangle$ are the two-qubit basis states that are not equal to $|LL\rangle$. Note, that this still damps out e.g. $|L0\rangle$. This is a bit of an odd physical process, but it matches the semantics of correlated loss meaning you always end up with both qubits being lost.

Overall, this should be straightforward to implement, once we finish #29 .

david-pl:
After discussion with @obigorda this isn't quite as simple:

We need to be able to set distinct probabilities for:
* losing any one atom when both atoms are initially in the qubit subspace ($PP \to PL$ or $PP \to LP$)
* losing both atoms simultaneously, when both are initially in the qubit subspace ($PP \to LL$)
* losing one atom when the other one has already been lost ($PL \to LL$ or $LP \to LL$),

where $P \in \\{I, X, Y, Z\\}$.

The channel described above is overly simplified in that it sets all these probabilities to the same value, which is incorrect.

PR #31, opened 2026-02-02
Implement LossyPauliWord, loss channel and reset channel

### LINKED ISSUE #29: Support loss in PP
Some drafts with details: #26 #28

david-pl:
Closes #29 and supersedes #28. 

Basically did the following here:

* Changed the `PauliWord` to be a implemented via a trait and refactored everything to work with that new trait (without breaking the API).
* Added a new `LossyPauliWord` implementing the same trait.
* Implemented `LossChannel` and `ResetLossChannel` traits for `PauliSum` to model actual noise processes.
* Implemented a new, custom truncation strategy, that works pretty much like `MaxPauliWeight`, but specifically for counting the number of `L`s in a word. These should be truncated at significantly lower numbers.
* ~~**Edit**: I went ahead and also included #30 in here, since it's a rather simple addition and it makes sense to review things in one go here.~~

A few things to note:

* The only place of the previous version that is impacted are the `Clifford + Rotation`: this is now a blanket impl for all words. That means, we also have to check for loss there and not do anything. For the standard `PauliWord`, this is done via a simple method that always return `false`, so it should just be eliminated on compilation. We could also separate the impls, but that would i) duplicate quite a bit of code and ii) make trait bounds in e.g. `PauliSum` uglier.
* The new API isn't super great, e.g. you'll have to write something like this, which isn't the nicest:
```rust
let strat = ppvm_runtime::strategy::MaxLossWeight(2);
let mut state: PauliSum<
    config::indexmap::ByteFxHashF64<
        1,
        MaxLossWeight,
        LossyPauliWord<[u8; 1], fxhash::FxBuildHasher>,
    >,
> = PauliSum::<
    config::indexmap::ByteFxHashF64<
        1,
        MaxLossWeight,
        LossyPauliWord<[u8; 1], fxhash::FxBuildHasher>,
    >,
>::builder()
.n_qubits(3)
.strategy(strat)
.build();
```
Maybe we can do something with type aliases here, but I have to see. In any case, Python bindings should work nicely still.


crates/ppvm-runtime/src/sum/rot1.rs
Roger-luo: the levi_civita function assumes 2-level system, but we are passing p_g here which is corresponding to the new PauliWord and could have more than original X,Y,Z operators?

Roger-luo: I guess if you try rx or rxx on the loss state this will trigger the behaviour?

crates/ppvm-tobc/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod propagate;
```
david-pl, line 1: @Roger-luo should this new crate really be part of this commit / PR? Is it used anywhere?

david-pl:
@Roger-luo this would be good to go from my side, but what about the tobc module? Do you want that added in?

PR #32, opened 2026-02-06
fix the test errors

Roger-luo:
@david-pl codex fixed these errors for me.

### DIFF
--- a/crates/ppvm-runtime/src/map/dashmap.rs
+++ b/crates/ppvm-runtime/src/map/dashmap.rs
@@ -158,7 +158,9 @@ where
         self.par_iter_mut().for_each(|mut entry| {
             let (k, v) = entry.pair_mut();
             if let Some((new_k, new_v)) = f(k, v) {
-                dest.insert(new_k, new_v);
+                dest.entry(new_k)
+                    .and_modify(|v| *v += new_v.clone())
+                    .or_insert(new_v);
             }
         })
     }
--- a/crates/ppvm-runtime/src/sum/rot1.rs
+++ b/crates/ppvm-runtime/src/sum/rot1.rs
@@ -1,6 +1,43 @@
 use crate::traits::*;
 use crate::{char::Pauli, config::Config, sum::PauliSum};
 
+impl<T: Config> RotationOneMapInsertClosure<T> for PauliSum<T>
+where
+    T::Coeff: std::ops::MulAssign,
+{
+    fn rotate_1_map_insert_closure(
+        k: &T::PauliWordType,
+        v: &mut T::Coeff,
+        axis: Pauli,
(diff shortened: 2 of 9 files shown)


crates/ppvm-runtime/src/sum/rot1.rs
```diff
@@ -13,6 +13,10 @@ where
         let (sin, cos) = theta.sin_cos();
         self.map_insert(|k, v| {
             let p_g = k.get(addr0);
+            if p_g as u8 > 3 || axis as u8 > 3 {
+                // Rotation is undefined on loss / non-Pauli symbols; keep term unchanged.
```
david-pl: Good catch on the rotations! I forgot to test non-Clifford stuff.

Nit pick here: the comment is a bit wrong. Rotations aren't undefined, the action is exactly what is done here: a lost qubit remains unchanged. That's the definition.

crates/ppvm-runtime/src/sum/rot2.rs
```diff
@@ -21,6 +21,10 @@ where
     ) {
         let (sin, cos) = theta.sin_cos();
         self.map_insert(|k, v| {
+            if k.get_lbit(a) || k.get_lbit(b) {
+                // Two-qubit rotations are undefined on loss; keep term unchanged.
+                return None;
```
david-pl: I'm not sure about this here. I think that if you apply a two-qubit rotation on two qubits, where just one is lost, then you're effectively doing a single-qubit rotation.

For example, $XX |0 L\rangle = |1L\rangle$.

crates/ppvm-runtime/src/map/dashmap.rs
```diff
@@ -158,7 +158,9 @@ where
         self.par_iter_mut().for_each(|mut entry| {
             let (k, v) = entry.pair_mut();
             if let Some((new_k, new_v)) = f(k, v) {
-                dest.insert(new_k, new_v);
+                dest.entry(new_k)
```
david-pl, line 161: I did not think about collision 😅

david-pl (changes requested):
I'll update the two-qubit rotation and ping you for another round of review. Other than that, good to go!

david-pl (approved):
@Roger-luo done, please have another look. I've made two changes:

1. If an axis is `Pauli::L`, we now panic. This is probably a user error and leads to undefined behavior since the model neglects coherences between the qubit states and `L`.
2. A two-qubit rotation where a single qubit has been lost now falls back to a one-qubit rotation. I've abstracted the logic that updates the map in the single-qubit case into a new trait for that. Note, that I've also briefly verified that this is the actual behavior when modeling a full three-level system and applying a two-qubit rotation on it.

PR #36, opened 2026-02-27
Add CI and some basic docs

david-pl:


david-pl:
@Roger-luo I think for the docs to deploy properly, we need to enable gh-pages on the repo. But I don't have the necessary permissions.

PR #49, opened 2026-04-11
perf(tableau): optimize MSD sampling ~32% faster

Roger-luo:
## Summary
- **MSD benchmark: 181µs → ~122.5µs (32.3% improvement)** via 7 targeted optimizations to the generalized stabilizer tableau measurement and gate paths
- Autotune experiment with 17+ iterations, 7 kept, documenting both successful and failed approaches with findings

## Optimizations (kept)
| # | Change | Improvement | Files |
|---|--------|-------------|-------|
| 1 | Precompute odd-phase destabilizer bitmask — replaces O(n) loop with `count_ones()` | +17% | `data.rs`, `measure.rs` |
| 2 | Bulk `BitArray::ZERO` in tableau reset | +4% | `data.rs` |
| 3 | `trailing_zeros` for lowest-bit extraction | +1.2% | `measure.rs` |
| 4 | XOR phase updates in Clifford inner loops | +2.5% | `gates/clifford.rs` |
| 5 | FxHashMap instead of std HashMap for u128 keys | +7.6% | `data.rs`, `measure.rs`, `Cargo.toml` |
| 6 | Fused PauliWord multiplication loops | +3.1% | `phase/mul.rs` |
| 7 | Skip normalize when no coefficients trimmed | +1.5% | `data.rs` |

### LLVM optimizes the PhasedPauliWord delegation chain extremely well
Three separate attempts to manually inline or optimize the basic Clifford gates (H, CZ) at the PhasedPauliWord level all **regressed by 4–28%**. The layered call chain `Tableau::cz → PhasedPauliWord::cz → PauliWord::cz` produces code that LLVM optimizes better than any hand-written equivalent we tried.

### XOR phase trick: works at Tableau level, hurts at PhasedPauliWord level
Adding 0 or 2 (mod 4) is mathematically equivalent to XOR with 0 or 2: `(phase + 2) % 4 == phase ^ 2`. We replaced `pw.add_phase(condition as u8 * 2)` with `pw.phase ^= (condition as u8) << 1` in the Tableau `CliffordExtensions` for **+2.5%**. However, the same transformation applied inside `PhasedPauliWord` methods was **4% slower** — it disrupts LLVM's register allocation and instruction scheduling for the inlined delegation chain. `add_phase` itself is already `#[inline(always)]` and `% 4` is already compiled to `& 3`. The issue is not that `add_phase` is slow — it's that changing code structure at the PhasedPauliWord level disturbs a fragile LLVM optimization equilibrium.

Note: a general-purpose `flip_sign(bool)` method can't replace `add_phase` because `add_phase` is also called with arbitrary phase values (1, 3) in PauliWord multiplication where XOR gives wrong results.

### FxHash vs SipHash for integer keys
FxHash for u128 keys gave the single biggest improvement (+7.6%). SipHash is designed for DoS resistance, not speed. For internal HashMaps with integer keys where DoS is not a concern, FxHash is ~5x faster.

### HashMap ordering is non-deterministic
One attempt to avoid cloning the coefficient Vec (by moving it into a HashMap and trimming from there) introduced a correctness bug: HashMap iteration order differs between instances, which broke coefficient ordering in idempotent measurements. The original clone approach preserves Vec ordering for Case B.

### Full experiment log
See `docs/autotune/msd-gates/log.md` for all 17+ iterations with detailed findings.

## Test plan
- [x] All 142 existing tests pass (`cargo test -p ppvm-tableau`)
- [x] Runtime tests pass (`cargo test -p ppvm-runtime`)
- [x] Back-to-back criterion benchmarks confirm each optimization
- [ ] Review benchmark numbers on CI hardware

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,33 +10,5 @@ ppvm-python/extended_tests/stim/stim-programs/
 
 # Managed by ion
 CLAUDE.md
-.agents/skills/ion-cli
-.claude/skills/ion-cli
-.agents/skills/receiving-code-review
-.claude/skills/receiving-code-review
-.agents/skills/requesting-code-review
-.claude/skills/requesting-code-review
-.agents/skills/using-git-worktrees
-.claude/skills/using-git-worktrees
-.agents/skills/brainstorming
-.claude/skills/brainstorming
-.agents/skills/dispatching-parallel-agents
-.claude/skills/dispatching-parallel-agents
-.agents/skills/executing-plans
-.claude/skills/executing-plans
-.agents/skills/finishing-a-development-branch
-.claude/skills/finishing-a-development-branch
-.agents/skills/subagent-driven-development
-.claude/skills/subagent-driven-development
-.agents/skills/systematic-debugging
-.claude/skills/systematic-debugging
-.agents/skills/test-driven-development
-.claude/skills/test-driven-development
-.agents/skills/using-superpowers
-.claude/skills/using-superpowers
-.agents/skills/verification-before-completion
-.claude/skills/verification-before-completion
-.agents/skills/writing-plans
-.claude/skills/writing-plans
-.agents/skills/writing-skills
-.claude/skills/writing-skills
+.agents/skills/
+.claude/skills/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -800,6 +800,7 @@ dependencies = [
  "bitvec",
  "bnum",
  "criterion",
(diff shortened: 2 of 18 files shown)


Roger-luo:
FYI. @david-pl I created a skill based on autoresearch to improve the performance, created the test to make sure no behaviour change during the optimization. This PR should identified all the minor performance improvements we can do before going SIMD or CUDA

PR #66, opened 2026-04-28
Implement circuit component and vm

### LINKED ISSUE #42: feat: vihaco circuit component backed by ppvm simulators
## Context

To integrate with the stellarscope virtualization platform, ppvm needs a [vihaco](https://github.com/QuEraComputing/stellarscope/tree/main/crates/vihaco) component that executes quantum circuit instructions by delegating to ppvm's simulation backends (Pauli Propagation runtime or Generalized Tableau).

A stub already exists in stellarscope at [`crates/stellarscope-circuit`](https://github.com/QuEraComputing/stellarscope/tree/main/crates/stellarscope-circuit): it defines 15 gate instruction variants (`X`, `Y`, `Z`, `H`, `S`, `T`, `RX/Y/Z`, `RXX/YY/ZZ`, `CNOT`, `CZ`, `U3`) but has no operands (no qubit indices or rotation angles) and a no-op `execute` that only logs. This issue is about building the real implementation in ppvm on top of it.

Prerequisite: QuEraComputing/stellarscope#51 (generic component macro support) must be merged first, since the ppvm circuit component will be generic over the simulation backend.

### 1. Complete the instruction set in `stellarscope-circuit`

The current instructions have no operands. They need to carry:
- Qubit indices (one or two per gate)
- Rotation angles for parametric gates (`RX`, `RY`, `RZ`, `RXX`, `RYY`, `RZZ`, `U3`)
- Noise channel instructions: `Depolarize1`, `Depolarize2`, `PauliChannel1`, `PauliChannel2`, `XError`, `YError`, `ZError` (matching what the STIM parser already handles in #44)
- `Measure` with a qubit index and a result-slot identifier

These additions should land in stellarscope, not ppvm — the stellarscope ISA is the shared contract.

### 2. Implement the ppvm circuit component

Add a new crate (e.g. `ppvm-circuit` or as part of `ppvm-vm`) that:
- Implements the vihaco `#[component]` macro on a struct generic over the simulation backend
- `execute` dispatches each instruction to the backend: Clifford + noise goes to the PP runtime or Tableau
- Measurement results are emitted as vihaco events so the machine can observe them

### 3. Define a minimal test machine

A simple machine (CPU + CircuitComponent) for integration testing, similar to `stellarscope-demo`. This does not need to be the production machine — its purpose is to verify the component can be wired into a vihaco machine and round-trip circuit execution correctly.

## Why this unblocks mixed-mode BFS/DFS (see #40)

Once circuits run through the vihaco VM, the VM execution model is the natural interface for multi-threaded DFS Pauli propagation:
- Each DFS path is a separate VM instance with its own simulation state
- Multi-threaded DFS = spawning one VM thread per path, each running the same circuit bytecode independently
- The BFS front-end decides when to branch; branches hand off a cloned VM state to a new thread

Without the circuit component, DFS multi-threading has no VM execution to parallelize.

david-pl:
Implements a vihaco component for circuit execution with instructions and a simple VM, finally putting the VM into ppvm.

Closes #42.

~~Requires https://github.com/QuEraComputing/stellarscope/pull/51 to be merged.~~
~~Requires https://github.com/QuEraComputing/stellarscope/pull/59 to be merged.~~
~~Requires https://github.com/QuEraComputing/vihaco/pull/3 to be merged.~~

CI fails because of dependency issues (the repo can't pull in vihaco, since that's still private).


Roger-luo:
I guess the dependency PR is this one now? https://github.com/QuEraComputing/vihaco/pull/3

Roger-luo:
question - how do I try out this right now as a human? do we have a CLI somewhere that reads the `*.sst` file or a `*.ssb` file?

crates/ppvm-vihaco/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+use crate::composite::PPVM;
+
+pub mod component;
+pub mod composite;
+pub mod instruction;
+pub mod measurements;
+pub mod message;
+mod syntax;
+
+pub fn run_file(path: &str) -> eyre::Result<PPVM> {
```
Roger-luo, line 41: I see. this answers my own question at the API level, I think we should have a `ppvm-cli` crate to provide a CLI interface via clap to execute these files given we don't have our compiler connected with this yet

crates/ppvm-vihaco/src/measurements.rs
```diff
@@ -0,0 +1,40 @@
+use eyre::Result;
+use vihaco::{Effects, observe};
+
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum MeasurementOutcome {
```
Roger-luo, line 16: can we leave a comment here about the decision we had on slack? e.g why we pick this over other alternatives etc.

crates/ppvm-vihaco/src/measurements.rs
```diff
@@ -0,0 +1,40 @@
...
+    Zero = 0,
+    One = 1,
+    Lost = 2,
+}
+
+pub type MeasurementResult = Vec<MeasurementOutcome>;
```
Roger-luo: maybe use `SmallVec` instead of `Vec` here? you will be using a `Vec<MeasurementResult>` later which is `Vec<Vec<...>>`, `SmallVec` should make your memory a bit more continuous without doing anything extra

crates/ppvm-vihaco/src/message.rs
```diff
@@ -0,0 +1,23 @@
...
+    QubitAndFloat(usize, f64),                      // RX, depolarize, ...
+    TwoQubit(usize, usize),                         // CX, CZ
+    TwoQubitAndFloat(usize, usize, f64),            // RXX, ...
+    QubitU3(usize, f64, f64, f64),                  // U3
+    QubitAndFloatArr3(usize, [f64; 3]),             // PauliError
+    TwoQubitAndFloatArr3(usize, usize, [f64; 3]),   // Correlated loss
+    TwoQubitAndFloatArr15(usize, usize, [f64; 15]), // TwoQubitPauliError
+
+    // batched instructions
+    QubitBatch(Vec<usize>),                                     // X, Y, Z, ...
```
Roger-luo: this should be `SmallVec` too, messages are sent between instructions, could be expensive

Roger-luo: what is batched instructions anyway?

crates/ppvm-vihaco/src/syntax.rs
```diff
@@ -0,0 +1,592 @@
...
+
+        let mut module = Module::default();
+        module.code = code;
+        module.strings = std::mem::take(&mut self.strings);
+        module.extra = info;
+        Ok(module)
+    }
+}
+
+type E<'src> = extra::Err<Simple<'src, char>>;
```
Roger-luo: what is this for? maybe too brief? `Err<'src>` might be better?

crates/ppvm-vihaco/src/syntax.rs
```diff
@@ -0,0 +1,592 @@
...
+    }
+}
+
+// ---- Everything below is 1:1 copy from Acamar with Acamar -> PPVM renaming ----
+
+/// A deferred branch whose target(s) couldn't be resolved at lowering time
+/// because the label may appear later in the function body. Patched in a
+/// second pass once all labels are known.
+#[derive(Debug)]
+enum BranchPatch {
```
Roger-luo, line 205: what is a "patch" this is a new concept to me... maye it came from @kaihsin ?

Roger-luo (commented):
some minor comments, and I think it would be nice to have a CLI wraps the run_file API as follows

```sh
ppvm run <file> # execute a file
ppvm dump <file> # dump a text file into binary
```

when we run a `<file>`, we will first assume this is binary, and try to read the PPVM magic number, if does not match our magic number, we will treat it as a text file and try to execute it as a text file.

not blocking, but I wonder how hard it is to support a step-by-step debugger and a REPL like

```sh
ppvm run # without a path
```

? which allows one to play instructions step-by-step

crates/ppvm-vihaco/src/lib.rs
david-pl: Do you want this to go into this PR? It sounds like it's enough effort to warrant a follow-up PR.

I was not aware this was a requirement so far. For the `ppvm dump <file>` we'll need byte code emission, that's also not done yet. Not sure if there's a convenient way to do this via vihaco, we basically don't customize any byte-code besides the magic number, so maybe there's a default way to do this?

crates/ppvm-vihaco/src/message.rs
david-pl: @Roger-luo see also #99. Batching is basically broadcasting over many qubits. One of the autotune optimizations you made was to batch common gates in the MSD circuit, e.g. `sqrt_x_batch`. That won another factor of about 2x in performance, so batching is genuinely a good idea. The state is a little weird right now, since there is no trait for this and only a small subset used in the MSD example is batched. These currently live in some impl block here: https://github.com/QuEraComputing/ppvm/blob/a4437d33f1481b80730b87e24efd4968a5f80ca5/crates/ppvm-tableau/src/gates/clifford.rs#L166

crates/ppvm-vihaco/src/syntax.rs
david-pl: That is literal copy-pasta from vihaco 😅

crates/ppvm-vihaco/src/syntax.rs
david-pl: See also https://github.com/QuEraComputing/ppvm/blob/a4437d33f1481b80730b87e24efd4968a5f80ca5/crates/ppvm-tableau/src/gates/clifford.rs#L166

If I understand correctly, it's basically a place holder for a label, which something can point to at a later point in the byte code. Patches are used to resolve this in a second pass.

crates/ppvm-vihaco/src/lib.rs
david-pl: @Roger-luo never mind my questions, I went and implemented it in this PR.

You can now do

```sh
cargo install --path crates/ppvm-cli
```

and then

```sh
ppvm run crates/ppvm-cli/examples/ghz.sst
ppvm dump crates/ppvm-cli/examples/ghz.sst  # dumps to ghz.ssb
```

I'll look into step-wise execution now.

david-pl: Can now do step-wise too. I re-used the CPU's breakpoint semantics, so you can run

```sh
ppvm debug <file>
```

and it will stop whenever it hits a breakpoint instruction. You can also do

```sh
ppvm debug -b <file>
```

to immediately halt execution at `pc = 0`, without having to add a break-point in the program file. You can then run things step-by-step interactively.

Roger-luo:
I'm gonna merge this first, we can come back to refactor if we are not happy with it, otherwise it's taking too long to review such a giant PR.

PR #74, opened 2026-05-17
fix(stim-parser): address proptest findings

Roger-luo:
## Summary

PR #73 added a proptest suite that surfaced two issues, both of which were worked around in the tests rather than fixed in the code. This PR removes the workarounds by fixing the underlying parser/AST.

## Finding 1 — stack overflow on deeply nested REPEAT

The chumsky grammar uses `recursive(...)` for REPEAT bodies, and the interpret pass recurses through `Repeat` bodies as well. Both pieces descend recursively at parse time, and the default thread stack (~2–8 MiB) overflowed somewhere past 24 nesting levels in debug builds.

**Fix:** run \`parse\` and \`parse_extended\` on a dedicated 16 MiB parser thread via \`std::thread::scope\`. Covers both the chumsky combinator stack and the interpret pass without touching the grammar itself.

\`proptest_parse.rs\`'s \`nested_repeats_never_panic\` cap is raised from 16 to 128.

## Finding 2 — \`Raw(MPad)\` never round-trips through \`parse_extended\`

\`parse_extended\` always lowers \`MPAD …\` into the typed \`ExtendedInstruction::MPad\`, so \`Raw(RawInstruction::MPad{..})\` and \`Raw(RawInstruction::Repeat{..})\` were states the parser never produced but the type system still allowed — causing runtime \`unreachable!()\` paths in \`prepare\` / \`executor\` and a workaround in the proptest generator.

**Fix:** introduce \`RawPassthrough\` (sub-enum of \`RawInstruction\` covering only \`Gate\` / \`Noise\` / \`Measure\` / \`Annotation\`) and change \`ExtendedInstruction::Raw(RawInstruction)\` to \`Raw(RawPassthrough)\`. The bad states are now compile-time unrepresentable.

Downstream impact:
- \`ExecError::Malformed\` dropped — it can no longer be reached.
- Two \`unreachable!()\` arms in the executor removed.
- \`proptest_ast.rs\` drops \`ext_raw_passthrough\` and generates \`Raw(RawPassthrough::*)\` directly across all four passthrough variants.
- The stale regression seed file is removed (the case it captured no longer compiles).
- \`tests/extended.rs\` pattern matches updated to the new variant.

## Test plan

- [x] \`cargo test --workspace --exclude ppvm-python-native\` — all green
- [x] \`pytest ppvm-python/test/generalized_tableau/test_stim.py\` — 31 pass
- [x] proptest_parse, proptest_roundtrip, proptest_ast all pass with the workarounds removed
- [x] \`nested_repeats_never_panic\` now exercises depths up to 128

### DIFF
--- a/crates/ppvm-stim/src/executor.rs
+++ b/crates/ppvm-stim/src/executor.rs
@@ -8,8 +8,8 @@ use std::fmt::Debug;
 
 use ppvm_runtime::prelude::*;
 use ppvm_tableau::prelude::*;
-use stim_parser::ast::{GateName, MeasureName, NoiseName, RawInstruction};
-use stim_parser::extended::{Axis, ExtendedInstruction, ExtendedProgram};
+use stim_parser::ast::{GateName, MeasureName, NoiseName};
+use stim_parser::extended::{Axis, ExtendedInstruction, ExtendedProgram, RawPassthrough};
 
 use crate::prepare::{ExecError, prepare};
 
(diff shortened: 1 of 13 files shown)


Roger-luo:
@copilot apply changes based on the comments in [this thread](https://github.com/QuEraComputing/ppvm/pull/74#pullrequestreview-4305861633)

PR #85, opened 2026-05-18
chore: address copilot review on #84 (CLA + hawkeye tooling)

Roger-luo:
## Summary

Follow-up to #84 addressing 3 of the 5 Copilot review comments:

- **`prek.toml`** — replace stale `cargo install hawkeye` comments and run the hawkeye hook via `mise exec --` so it works regardless of whether the shell is mise-activated.
- **`CLA.md`** — the intro promised commit-signing instructions but the referenced section only describes PR submission. Re-worded the intro to match.
- **`.github/workflows/cla.yml`** — `issue_comment` fires on issue comments too. Added `github.event.issue.pull_request != null` to the trigger guard so the job only runs on PR comments.

## Intentionally not changed

- **Drop `actions: write`** — `contributor-assistant/github-action` uses it to re-run the workflow after a signature so the status check refreshes; removing it would break the sign-then-merge flow.
- **Commit an empty `cla-signatures.json`** — the action creates the file on first signature with its expected schema; pre-seeding could conflict.

## Test plan

- [ ] CI passes (Rust + Python + License headers + CLA Assistant).
- [ ] Confirm the CLA Assistant workflow still fires on this PR's `pull_request_target: opened` (since the trigger guard change preserves that branch).
- [ ] After merge, sanity-check the prek hook locally: `prek run hawkeye`.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -17,7 +17,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: CLA Assistant
-        if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
+        if: |
+          (github.event_name == 'issue_comment' &&
+           github.event.issue.pull_request != null &&
+           (github.event.comment.body == 'recheck' ||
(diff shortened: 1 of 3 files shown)


Roger-luo:
I have read the CLA Document and I hereby sign the CLA

PR #94, opened 2026-05-20
feat(runtime): U(1)-conserving Pauli propagation helpers

AlexSchuckert:
## Summary

- Adds `exchange(a, b, θ)` for `exp[−i θ/2 (X_a X_b + Y_a Y_b)]` on `PauliSum`.
- Adds `xyzz(a, b, θ_xy, θ_zz)` combining XY exchange with a `ZZ` interaction.
- Adds a Python-level `apply_u1_trotter_step(edges, theta_xy, theta_zz, fields_z)` helper.
- Intended for XY and Heisenberg-style z-magnetization-conserving spin models.
- Both `exchange` and `xyzz` go through the existing `RotationTwo` primitives (`rxx`/`ryy`/`rzz`); no new operator basis, no new backend.

The new trait `U1Conserving<T: Config>` lives at `crates/ppvm-runtime/src/traits/branch/u1.rs`; the `PauliSum` impl at `crates/ppvm-runtime/src/sum/u1.rs`. Exposed through the existing `create_interface!` PyO3 macro so every config variant picks them up. Python wrapper methods on `RotationsMixin` (`exchange`, `xyzz`) and `PauliSum.apply_u1_trotter_step` round out the surface.

## Conservation under truncation

The trait and Python docstrings document the precise guarantee: `{I, Z}`-only Pauli observables (`Σ_i Z_i`, `Σ_{i<j} Z_iZ_j`, …) are conserved through propagation up to per-gate floating-point precision (~1e-15 per gate, accumulating linearly), as long as the truncation cutoff sits comfortably below the conserved-coefficient magnitude (`~1`) and well above the per-gate ε floor. Aggressive `CoefficientThreshold` settings close to 1, or very long circuits, can break this; the hardening tests below pin the safe regime.

## Tests

**Rust** (11 tests in `crates/ppvm-runtime/src/sum/u1.rs::tests`):
- `exchange` == `rxx` then `ryy` on every two-qubit Pauli, at multiple angles.
- `xyzz` == `exchange` then `rzz`.
- Zero-angle identity for both gates.
- `Σ Z_i` and `Σ_{i<j} Z_iZ_j` preservation through `exchange`/`xyzz`.
- 4 hardening tests asserting truncation-robust `Σ Z` conservation under `CoefficientThreshold(0.5)`, `MaxPauliWeight(1)`, and a manual Trotter sweep.

**Python** (16 tests in `ppvm-python/test/test_u1.py`):
- API surface: `exchange`, `xyzz`, `apply_u1_trotter_step` all exist and run.
- Match against manual gate application.
- Per-edge couplings and per-edge angle sequences.
- 3 hardening tests for total-Z and full-pair ZZ conservation under aggressive truncation.

## Test plan
- [x] `cargo test --workspace`
- [x] `uv run --project ppvm-python --group dev pytest ppvm-python/test/`
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `ruff format --check`, `ruff check`


david-pl:
@AlexSchuckert the changes here look fine, since they are mostly small additions. Just to clarify, however, could this also be achieved by allowing control over when to truncate from the python side?

AlexSchuckert:
Thanks! Yes I think so, that would be equivalent.

AlexSchuckert:
Superseded by #100 (`deferred-truncate`). That branch replaces `exchange` / `xyzz` with a generic `truncate: bool = True` kwarg on every gate plus an explicit `PauliSum.truncate()` — the user controls *when* the configured strategy fires, which subsumes the U(1)-conserving fusion this PR was doing as a special case (and generalises to any commuting-gate composition).

PR #98, opened 2026-05-27
Adaptive Pauli-Lindbladian shim + adaptive-evolution demo

AlexSchuckert:
## Summary

Adds a Rust shim for **direct, Trotterless Heisenberg-picture time evolution of an observable under a generic Pauli-Lindbladian** (Hermitian Pauli Hamiltonian `H = Σ c_i P_i` + Hermitian Pauli jump operators `L_k` with rates `γ_k`), plus a Python driver demonstrating an adaptive Pauli-basis integrator built on it.

### `LindbladSpec` (Rust, `crates/ppvm-python-native/src/lindblad.rs`)
A precompiled Lindbladian exposing three primitives over bit-packed Pauli words:
- `action(p)` — `L*(p)` for a single Pauli string;
- `leakage(basis, coeffs[, protected])` — the off-basis component of `L*(Σ c_j p_j)` (drives adaptive basis growth);
- `generator(basis)` — the Lindbladian restricted to `basis` as sparse COO triplets (wrapped into a scipy CSC matrix on the Python side).

Hot path: a fused per-byte commutator computing both the XOR product and its phase in one pass; per-qubit active-site index so only terms whose support meets `supp(p)` are visited; a `DashMap` action cache (the analogue of `@lru_cache` in the prototype); `rayon`-parallel basis loops. Numpy `uint8` (`0=I,1=X,2=Z,3=Y`) I/O end-to-end, so the driver never builds per-row strings on the hot path.

### `ppvm.Lindbladian` (Python wrapper, `ppvm-python/src/ppvm/lindblad.py`)
Thin wrapper with both ndarray-native (`action_arr`/`leakage_arr`/`generator_arr`) and string-keyed (`action`/`leakage`/`generator`) APIs.

### Demo (`ppvm-python/demo/lindblad_adaptive.py`)
End-to-end adaptive integrator: at each step it measures the leakage out of the current basis, adds the leaked strings above `add_tol`, advances `exp(dt · M)` (exact in dt within the basis — no Trotter error) via `scipy.sparse.linalg.expm_multiply`, and prunes below `drop_tol` (a protected set is never dropped). `--predictor_corrector 1` enriches the basis with the predictor's second-hop strings and redoes the step, lifting the dt-error from `O(dt²)` to `O(dt³)`. Reports `C_k(t)` and a discarded-weight a-posteriori error monitor.

### Packaging
`import ppvm` no longer requires SciPy: SciPy is imported lazily inside `generator()`/`generator_arr()` (the only methods that need it). `numpy` is declared as a runtime dependency (used by the wrapper).

## Test plan
- [x] `ppvm-python/test/test_lindblad.py` cross-checks `action`/`leakage`/`generator`/protected-set against an independent dense Pauli-multiplication-table reference (XY+Z-dephasing and TFIM+X-dephasing); 4 tests pass.
- [x] `cargo fmt --check` and `cargo clippy` clean on `ppvm-python-native`.
- [x] `ruff check` clean on the wrapper and test.
- [x] `uv lock --check` consistent.
- [x] Demo runs end to end (plain and `--predictor_corrector`).


Roger-luo:
@copilot fix the license header for Alex :-)

Roger-luo:
@AlexSchuckert wondering if we can have some information about how you verify the correctness of this to help reviewing?

crates/ppvm-lindblad/src/expm.rs
```diff
@@ -0,0 +1,405 @@
...
+
+/// Compute `exp(t · A) · b` for a real square sparse matrix `A` in CSR
+/// form. Allocates a fresh `Vec<f64>` of length `A.n`.
+///
+/// The algorithm is Al-Mohy & Higham (2011) Algorithm 3.2: pick
+/// `(m, s)` from the precomputed `θ_m` table to minimise SpMV count,
+/// then iteratively apply a degree-`m` Horner-form Taylor polynomial
+/// `s` times. Inside each Horner sweep we terminate early once two
+/// successive Taylor terms have norm below `opts.tol · ‖F‖`.
+pub fn expm_multiply(a: &CsrMatrix, t: f64, b: &[f64], opts: ExpmOpts) -> Vec<f64> {
```
Roger-luo: I'm wondering if this exists in a rust crate already?

crates/ppvm-lindblad/src/expm.rs
```diff
@@ -0,0 +1,405 @@
...
+    (26, 2.64),
+    (27, 2.86),
+    (28, 3.08),
+    (29, 3.31),
+    (30, 3.54),
+];
+
+/// Compressed Sparse Row matrix (square, `n × n`).
+#[derive(Clone, Debug)]
+pub struct CsrMatrix {
```
Roger-luo: same, I'd prefer if we can use an existing crate instead of writing our own CSRmatrix from scratch

Roger-luo: also note that we have a sparsevec made from scratch in PPVM-tableu, might worth considering if there are ways to unifying them (at least the interface) like how people do it in Julia.

cc: @david-pl

crates/ppvm-lindblad/examples/spmv_scaling.rs
```diff
@@ -0,0 +1,65 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Tiny SpMV scaling check: builds a tridiagonal CSR matrix of a target
+//! `nnz`, times `spmv_parallel` repeatedly inside rayon pools of varying
+//! thread counts, and prints the wall time and speedup. Useful to confirm
+//! the SpMV path itself scales before debugging higher-level paths.
+
+use ppvm_lindblad::CsrMatrix;
+use std::time::Instant;
```
Roger-luo: can we actually turn this into benchmark like other crates instead of using `std::time` to measure the performance? it will be cleaner to use the same setup as other crates like those in ppvm-runtime

ppvm-python/demo/lindblad_adaptive.py
```diff
@@ -0,0 +1,260 @@
...
+    target_coeff_arr = coeff.copy()
+    protected_arr = basis_arr.copy()  # never dropped, never emitted as leakage
+    protected_keys = {row.tobytes() for row in protected_arr}
+    index = {row.tobytes(): i for i, row in enumerate(basis_arr)}
+    Ck0[m] = float(np.dot(coeff, coeff))
+    norm_target = np.sqrt(Ck0[m])
+
+    L_op.clear_cache()  # action cache is per-k since each k starts a fresh basis
+
+    for nt in range(c.steps + 1):
```
Roger-luo: I understand this might work... but it would be nice to split these nested loops into functions to make things easier to read

ppvm-python/demo/lindblad_adaptive.py
```diff
@@ -0,0 +1,260 @@
...
+    for j, xj in enumerate(x):
+        coef = float(np.cos(kk * xj))
+        if abs(coef) > c.drop_tol:
+            target_rows.append(zterm_codes(j))
+            target_coeffs.append(coef)
+    basis_arr = np.array(target_rows, dtype=np.uint8)
+    coeff = np.array(target_coeffs, dtype=float)
+    target_arr = basis_arr.copy()
+    target_coeff_arr = coeff.copy()
+    protected_arr = basis_arr.copy()  # never dropped, never emitted as leakage
```
Roger-luo: these copy looks suspicious

ppvm-python/demo/lindblad_pc_scaling.py
```diff
@@ -0,0 +1,184 @@
...
+            tau_add,
+            protected=protected,
+            parallel_threshold=0,
+            num_threads=num_threads,
+        )
+        times.append(time.perf_counter() - t0)
+    return times, len(basis)
+
+
+def main():
```
Roger-luo: it would be better if we can have all the examples written in jupytext instead of CLI?

ppvm-python/src/ppvm/lindblad.py
```diff
@@ -0,0 +1,330 @@
...
+# Lookup table mapping code -> ASCII byte for vectorised string output.
+_CODE_TO_ASCII = np.array([ord("I"), ord("X"), ord("Z"), ord("Y")], dtype=np.uint8)
+
+# A jump operator is either a Hermitian Pauli (single string) or a complex
+# linear combination of Pauli strings.
+PauliLincomb = Iterable[tuple[str, complex]]
+JumpSpec = Union[tuple[str, float], tuple[PauliLincomb, float]]
+
+
+def string_to_codes(s: str, n_qubits: int) -> np.ndarray:
```
Roger-luo: I understand what this does but what's the use case?

Roger-luo: if it's only used to build Linbaldian, I would keep it module private by adding `__` prefix instead of expose it to downstream user

ppvm-python/test/test_lindblad.py
```diff
@@ -0,0 +1,723 @@
...
+
+    rng = np.random.default_rng(99)
+    for _ in range(5):
+        p = _random_pauli_str(rng, L)
+        got_a = L_simple.action(p)
+        got_b = L_lincomb.action(p)
+        for k in set(got_a) | set(got_b):
+            assert got_a.get(k, 0.0) == got_b.get(k, 0.0), (
+                f"lincomb fast path mismatch at p={p!r}: {got_a} vs {got_b}"
+            )
```
Roger-luo: a 700 loc python file is a hard reject

ppvm-python/pyproject.toml
```diff
@@ -10,6 +10,7 @@ requires-python = ">=3.10"
 dependencies = [
     "bloqade-circuit>=0.14.1",
     "kirin>=0.0.15",
+    "numpy>=1.26",
     "ppvm-python-native",
 ]
```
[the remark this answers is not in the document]
Roger-luo: this is valid concern, please address

Roger-luo (changes requested):

crates/ppvm-lindblad/src/expm.rs
AlexSchuckert: No Rust crate provides `expm_multiply` in the matrix-free action form (Al-Mohy & Higham, scaling-and-squaring on `A · v`). The `expm` crate exists but only handles dense matrices via the Padé approximation — not useful for our sparse Heisenberg-picture generator. SciPy and Julia (`KrylovKit` / `Expokit` ports) both have it; we follow SciPy's `expm_multiply` algorithm and `θ_m` table directly. It's ~200 LOC of self-contained numerics, verified by `expm_serial_matches_parallel` and cross-checked against `scipy.sparse.linalg.expm_multiply` in `test_pc_step_rust_matches_python_pc`. Happy to swap if a maintained crate shows up.

crates/ppvm-lindblad/src/expm.rs
AlexSchuckert: Surveyed `sprs`, `nalgebra-sparse`, `russell_sparse`, and `faer-rs`. All four provide CSR + serial SpMV, but **none** ship a rayon-parallel SpMV with the nnz-threshold gating we use here (small matrices stay single-threaded; only above `ExpmOpts::parallel_threshold` do we pay the rayon overhead). Pulling any of them in would mean wrapping their CSR to add the parallel-with-threshold path — net code surface unchanged, plus an extra dep.

On unifying with `ppvm-tableau::SparseVector`: that trait describes a sparse **vector** (single column, keyed by index), not a sparse matrix. Our `CsrMatrix` is the latter — different abstractions. The Julia analogy would be `AbstractSparseArray` shared between `SparseVector` and `SparseMatrixCSC`, but the shared surface for the two types here (a 1D coefficient store vs 2D row/col indices for SpMV) is too thin to make a common trait worth it.

Happy to revisit if you have a specific crate in mind. cc: @david-pl

AlexSchuckert:
> @AlexSchuckert wondering if we can have some information about how you verify the correctness of this to help reviewing?

I'm using several physics examples in test/lindblad. There it essentially builds the exact evolution using dense matrices. Of course I can't fully exclude weird edge cases, but I think these examples are non-trivial.

crates/ppvm-lindblad/src/expm.rs
david-pl: Also, one specific requirement for the sparse vector (which is why we have our own version, since no crate has this AFAIK): we need support for arbitrary bit-width indices since we need to support indices up to $2 ^ n$.

AlexSchuckert: Good point — and that gap actually reinforces keeping them separate. `ppvm-tableau::SparseVector` is keyed by bit-width-arbitrary indices (up to $2^n$, i.e. exponential-in-qubit-count), so it needs a generic index type. `ppvm-lindblad::CsrMatrix` is keyed by basis position — bounded by the Pauli-string basis size (typically $10^5$–$10^6$), so `u32` is comfortable. Any common abstraction would have to be generic over the index type, which lands us back to roughly the surface `ppvm-tableau::SparseVector` already exposes — without any of the matrix structure (`row_ptr`, parallel SpMV, etc.) that `CsrMatrix` needs.

So the two address different scaling regimes with deliberately different storage. Happy to revisit if a unifying abstraction surfaces; for now they stay as separate custom types.

AlexSchuckert: Update — went ahead and did the swap. `9fd6ef5` replaces the custom `CsrMatrix` with `sprs::CsMatI<f64, u32, usize>` for storage, keeping a thin `spmv_parallel` adapter (the one piece no crate provides). 

Type choice note: the default `sprs::CsMat<f64>` uses `usize` column indices (8 bytes); benchmarking showed that causes a 17–32% regression at bandwidth-bound matrix sizes (8M+ nonzeros) from doubled per-nonzero memory traffic. `CsMatI<f64, u32, usize>` keeps the original 4-byte indices and matches the previous performance within ±5% noise across all matrix sizes I tested (100k → 8M).

Net: ~25 LOC custom CSR code removed, storage type now from the standard ecosystem. All Rust + Python tests pass. The `expm_multiply` algorithm stays ours per the other thread.

crates/ppvm-lindblad/benches/spmv_scaling.rs
```diff
@@ -0,0 +1,66 @@
...
+                );
+            });
+        }
+
+        group.finish();
+    }
+}
+
+criterion_group!(benches, bench_spmv_scaling);
+criterion_main!(benches);
```
Roger-luo: can we post in the PR comment about the result of this benchmark for future reference?

crates/ppvm-lindblad/src/expm.rs
```diff
@@ -0,0 +1,400 @@
...
+    let indptr = indptr_raw.raw_storage();
+    let indices = m.indices();
+    let data = m.data();
+    for (i, yi) in y.iter_mut().enumerate() {
+        let mut sum = 0.0;
+        for k in indptr[i]..indptr[i + 1] {
+            sum += data[k] * x[indices[k] as usize];
+        }
+        *yi = sum;
+    }
```
Roger-luo: can we verify if this gets correctly vectorized after compilation?

crates/ppvm-python-native/src/lib.rs
```diff
@@ -1,10 +1,18 @@
 // SPDX-FileCopyrightText: 2026 The PPVM Authors
 // SPDX-License-Identifier: Apache-2.0
 
+// mimalloc returns freed pages to the kernel more aggressively than the
+// default system allocator. This materially reduces peak RSS for the
+// allocation-heavy adaptive-Pauli paths (leakage + generator each
+// allocate hundreds of MB of transient Vec data per pc_step).
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
```
Roger-luo: This looks suspicious. I think we need more explanation of why we are adding this custom allocator and how much improvement this gives compared to not having it.

Ideally, the initial draft PR should not contain this type of optimization; we should add that in a separate PR. I would decline to merge this draft PR as a single unit.

It would be better to start with the basic setup and have that reviewed first, then submit a separate PR focusing on the different optimizations. This avoids forcing the reviewer to go through a giant 10,000-line PR.

crates/ppvm-lindblad/src/expm.rs
```diff
@@ -0,0 +1,400 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
```
Roger-luo, line 1: If you're adding this as an internal implementation, it would be nice to add this as a separate crate instead of within ppvm-liblblod. I imagine this will be useful for other Rust developers as well, and it will also help clean up the dependency a little bit.

It would be nice to first submit a separate PR that creates a single crate that implements sparse matrix and exponential.

crates/ppvm-lindblad/src/lib.rs
```diff
@@ -0,0 +1,1169 @@
...
+
+/// Maximum number of qubits supported by [`Word`] (= `8 · W_U64 · sizeof(u64)`).
+pub const MAX_QUBITS: usize = 64 * W_U64;
+
+/// The Pauli-word storage type used throughout this crate.
+///
+/// `[u64; 2]` covers up to 128 qubits; the `FxBuildHasher` matches the
+/// hash used by the `FxHashMap` keys we wrap with; `REHASH=true` means
+/// `set()` keeps the cached hash in sync.
+pub type Word = PauliWord<[u64; W_U64], FxBuildHasher, true>;
```
Roger-luo, line 79: Is there a reason why our implementation has a hard-coded number of qubits here?

crates/ppvm-lindblad/src/lib.rs
```diff
@@ -0,0 +1,1169 @@
...
+
+/// Errors raised when constructing a [`LindbladSpec`].
+#[derive(Debug, Clone)]
```
Roger-luo: Let's put all the error handling in a separate file, so we have a clean lib.rs file.

crates/ppvm-lindblad/src/lib.rs
```diff
@@ -0,0 +1,1169 @@
...
+/// `Send + Sync` wrapper around `*mut T` for parallel scatter writes from
+/// rayon tasks into pre-allocated arrays. Caller must guarantee that the
+/// writes target disjoint indices.
+///
+/// The field is intentionally private so closures that use [`Self::write`]
+/// capture the whole struct (which is `Send + Sync`) rather than the
+/// inner `*mut T` (which is not) — Rust 2021's disjoint-field capture
+/// would otherwise peel the wrapper.
+#[derive(Clone, Copy)]
+struct SendPtr<T>(*mut T);
```
Roger-luo: This is an interesting approach. Does this have an existing Rust standard library functionality that we can reuse?

We do not accept any unsafe code in this crate in general. If we have to use unsafe, it must be separated and isolated in a separate crate.

crates/ppvm-lindblad/src/lib.rs
```diff
@@ -0,0 +1,1169 @@
...
+    ///
+    /// Lifts the per-step truncation error from `O(dt²)` (single-hop) to
+    /// `O(dt³)`. Strings in `protected` are never added to the basis as
+    /// leakage targets — typically the observable's support, which the
+    /// caller wants tracked exactly.
+    ///
+    /// `num_threads`, when set, runs the entire step inside a freshly built
+    /// rayon thread pool of that size — useful for benchmarking parallel
+    /// scaling. When `None`, the global rayon pool is used.
+    #[allow(clippy::too_many_arguments)]
```
Roger-luo: We should not work around Clippy errors like this. If Clippy warns you about too many arguments, this is a code smell of an incorrect abstraction being created here.

Ideally, you would create an object that contains configurations such as:
1. Drop tolerance
2. Tau add
3. Other relevant options

Then, you can feed that object in with the Hamiltonian and Lindblad stuff.

crates/ppvm-lindblad/src/lib.rs
```diff
@@ -0,0 +1,1169 @@
...
+        } else {
+            self.pc_step_inner(basis, coeffs, dt, tau_add, drop_tol, protected, opts)
+        }
+    }
+
+    /// Same as [`Self::pc_step`] but also returns a per-phase timing
+    /// breakdown (microseconds), for profiling parallel scaling and hot
+    /// spots. Output: `(leakage1, expand1, gencsr1, expm1, leakage2,
+    /// expand2, gencsr2, expm2)`.
+    #[allow(clippy::too_many_arguments)]
```
Roger-luo: Same here.

ppvm-python/src/ppvm/lindblad.py
```diff
@@ -0,0 +1,193 @@
...
+_PAULI_CODE = {"I": 0, "X": 1, "Z": 2, "Y": 3}
+# Lookup table mapping code -> ASCII byte for vectorised string output.
+_CODE_TO_ASCII = np.array([ord("I"), ord("X"), ord("Z"), ord("Y")], dtype=np.uint8)
+
+
+def string_to_codes(s: str, n_qubits: int) -> np.ndarray:
+    """Encode a Pauli string ``"IXYZ..."`` as a length-``n_qubits`` uint8 array."""
+    if len(s) != n_qubits:
+        raise ValueError(f"Pauli string {s!r} has length {len(s)} != n_qubits {n_qubits}")
+    return np.array([_PAULI_CODE[c] for c in s], dtype=np.uint8)
```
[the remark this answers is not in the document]
AlexSchuckert: Resolved: `_string_to_codes` now strips `_` and raises a `ValueError` naming the bad character and expected alphabet (matching the Rust parser), instead of letting a `KeyError` propagate.

crates/ppvm-python-native/src/lindblad.rs
```diff
@@ -0,0 +1,616 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Rust shim for direct Pauli-Lindbladian time evolution on an adaptive
+//! Pauli-string basis. See `xy-experiments/main_k_adaptive.py` for the
+//! Python driver that owns the linear algebra and basis adaptation.
```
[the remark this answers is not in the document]
AlexSchuckert: Resolved: the stale `xy-experiments/main_k_adaptive.py` reference is gone; the in-tree driver is `ppvm-python/demo/lindblad_adaptive.py`.

crates/ppvm-python-native/src/lindblad.rs
```diff
@@ -0,0 +1,616 @@
...
+
+        let mut merged: FxHashMap<Word, f64> = FxHashMap::default();
+        for m in local {
+            for (k, v) in m {
+                if in_basis.contains_key(&k) || protected_set.contains_key(&k) {
+                    continue;
+                }
+                *merged.entry(k).or_insert(0.0) += v;
+            }
+        }
```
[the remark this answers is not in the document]
AlexSchuckert: Addressed in b07f75de: `leakage`/`leakage_with_prune` now fold into thread-local `FxHashMap`s and merge via Rayon `reduce`, so the `Vec<HashMap>` of per-element intermediates is no longer materialized. Keys and summed values are identical (FP reduction order can differ at ~1e-16, well within the test tolerances).

crates/ppvm-python-native/src/lindblad.rs
```diff
@@ -0,0 +1,616 @@
...
+                self.n_qubits
+            )));
+        }
+
+        let basis_words = decode_basis(&basis_view, self.n_qubits)?;
+        let index: FxHashMap<Word, u32> = basis_words
+            .iter()
+            .enumerate()
+            .map(|(i, w)| (*w, i as u32))
+            .collect();
```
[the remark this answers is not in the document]
AlexSchuckert: Resolved: the `generator()` binding (and the pc_step paths) now call `assert_basis_unique`, which returns a `ValueError` naming the duplicate row indices before any matrix is built.

ppvm-python/pyproject.toml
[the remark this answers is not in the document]
AlexSchuckert: Resolved: `scipy>=1.13` is now an explicit entry in the `dev` dependency group in `ppvm-python/pyproject.toml` (runtime `ppvm` stays scipy-free).

crates/ppvm-lindblad/src/expm.rs
AlexSchuckert: Resolved: the matrix-exponential action is now provided by the external `quspin-expm` crate (QuSpin-rust), used matrix-free. The hand-rolled implementation has been removed; `expm.rs` is reduced to just the `(m, s)` selection table.

crates/ppvm-lindblad/examples/spmv_scaling.rs
AlexSchuckert: Resolved: the `examples/spmv_scaling.rs` timing example has been removed -- the real-CSR SpMV it exercised no longer exists (evolution is matrix-free via `quspin-expm`).

ppvm-python/demo/lindblad_adaptive.py
AlexSchuckert: Addressed: the demo is now factored into helper functions -- `build_xy_dephasing`, `init_mode`, `add_leakage_to_basis`, `prune_and_cap`, and `run_mode` -- rather than inline nested loops.

ppvm-python/test/test_lindblad.py
AlexSchuckert: Resolved: the monolithic `test_lindblad.py` has been split into `ppvm-python/test/lindblad/` (`test_action_generator.py`, `test_adaptive_pc.py`, `test_non_hermitian_jumps.py`, `test_pc_step_rust.py`, plus `_helpers.py`).

crates/ppvm-lindblad/benches/spmv_scaling.rs
AlexSchuckert: Obsolete: the `spmv_scaling` benchmark has been removed -- it benchmarked the now-deleted real-CSR SpMV. The matrix-exponential action is now provided by the external `quspin-expm` crate.

crates/ppvm-lindblad/src/expm.rs
AlexSchuckert: Obsolete: the hand-rolled scaling-and-squaring inner loop has been removed; the matrix-exponential action is now provided by the external `quspin-expm` crate.

crates/ppvm-python-native/src/lib.rs
AlexSchuckert: Addressed: the global allocator has been removed from this PR (f14edafc) and moved to its own PR -- #129 -- so it can be reviewed and benchmarked independently, per your suggestion.

crates/ppvm-lindblad/src/expm.rs
AlexSchuckert: Addressed: rather than a first-party crate, the matrix-exponential is now provided by the external `quspin-expm` crate (QuSpin-rust). `expm.rs` is reduced to just the `(m, s)` selection table, so the dependency surface is much smaller.

crates/ppvm-lindblad/src/lib.rs
AlexSchuckert: Addressed in b07f75de: added docs explaining the fixed-width `[u64; W_U64]` Pauli-word storage and `MAX_QUBITS = 64 * W_U64` -- chosen for cache-friendly O(1) Pauli ops; construction beyond it errors with `Error::TooManyQubits`.

crates/ppvm-lindblad/src/lib.rs
AlexSchuckert: Addressed in b07f75de: the `Error` enum and its `Display`/`Error` impls moved to `crates/ppvm-lindblad/src/error.rs`; the public path `ppvm_lindblad::Error` is unchanged.

crates/ppvm-lindblad/src/lib.rs
AlexSchuckert: Resolved: all `unsafe` has been removed from `ppvm-lindblad`. The `SendPtr` parallel-scatter helper is gone along with the CSR generator it served -- evolution is now matrix-free via `quspin-expm`.

crates/ppvm-lindblad/src/lib.rs
AlexSchuckert: Addressed in b07f75de: introduced `PcStepConfig { tau_add, drop_tol, num_threads }`; `pc_step`/`pc_step_timed` (and their inner helpers) now take it, and the `#[allow(clippy::too_many_arguments)]` has been removed. The Python-facing kwargs are unchanged -- the binding builds the config internally.

crates/ppvm-lindblad/src/lib.rs
AlexSchuckert: Addressed in b07f75de: covered by the same `PcStepConfig` refactor.

AlexSchuckert:
Superseded by #178, which carries the final API (PcStepConfig, error module, sprs feedback all incorporated) plus the momentum-sector evolution and the full benchmark validation.

PR #100, opened 2026-05-29
feat(python): per-gate truncate kwarg + explicit PauliSum.truncate()

### LINKED ISSUE #63: Make truncation after every step optional in PP python bindings
Right now, we truncate after every gate / noise channel. This is a good default as it's usually best for performance. However, there might be cases where users need control over this behavior.

cc @AlexSchuckert

AlexSchuckert:
## Summary

Replaces the per-gate auto-truncate with an opt-out: every gate method on `PauliSum` / `LossyPauliSum` (and `GeneralizedTableau`, for signature symmetry) gains a `truncate: bool = True` kwarg, and `PauliSum` gains a top-level `truncate()` method. Default behaviour is **exactly** unchanged; existing user code is unaffected.

```python
ps.rxx(a, b, theta, truncate=False)
ps.ryy(a, b, theta, truncate=False)
ps.truncate()        # one strategy run at the end of the commuting pair
```

## Why

Supersedes #94 (`u1-conserving`). That branch added a dedicated `exchange` (and `xyzz`) gate whose only job was to call `rxx + ryy` back-to-back in Rust so the PyO3-level auto-truncate fired once instead of twice — the U(1) charge structure is preserved by the single PyO3 truncate, not by `rxx + ryy`'s two truncates.

Letting the Python caller control `truncate` directly subsumes `exchange` and generalises to any sequence of commuting gates (`xyzz = rxx + ryy + rzz`, longer fusions, mixed Clifford-rotation blocks). No new dedicated gates needed, no new Rust traits, smaller API surface.

## Design

- The PyO3 `PauliSum` interface auto-truncate now lives behind `if truncate { self.inner.truncate(); }` on every gate. The kwarg defaults to `true`, so callers that don't pass it see no change.
- A new `PauliSum.truncate()` calls the configured strategy explicitly.
- The Python mixins (`CliffordMixin`, `RotationsMixin`, `CliffordExtensionMixin`, `NoiseMixin`, `LossMixin`) forward `truncate=truncate` to the PyO3 layer.
- `GeneralizedTableau`'s PyO3 binding also accepts the kwarg and silently ignores it — the tableau backend doesn't auto-truncate, so there's nothing to defer, but the parallel signature lets the mixins serve both backends uniformly.

## Tests

`ppvm-python/test/test_truncate_kwarg.py` (7 cases):

- `test_default_truncate_kwarg_is_true` — backward-compat: no kwarg ≡ explicit `truncate=True`.
- `test_truncate_false_then_explicit_matches_implicit` — at a threshold that doesn't actually drop anything, deferred ≡ immediate.
- `test_truncate_false_keeps_intermediate_terms_alive` — at a threshold where the immediate strategy *would* drop intermediates, `truncate=False` keeps them visible in the post-gate state.
- `test_truncate_false_pair_reproduces_old_exchange` — the new idiom produces the same state as what `exchange` did internally.
- `test_deferred_truncate_preserves_total_z_with_loose_threshold` — sanity: rxx+ryy chain with deferred truncate conserves Σ Z exactly.
- `test_truncate_method_drops_below_threshold_keys` — explicit `truncate()` is wired through and runs the configured strategy.
- `test_noise_channels_accept_truncate_kwarg` — the kwarg works on `pauli_error` too, not only on rotations.

## Test plan
- [x] `cargo test -p ppvm-runtime --lib` (78 pass)
- [x] `uv run --project ppvm-python --group dev pytest ppvm-python/test/` (109 pass)
- [x] `cargo fmt`, `cargo clippy --no-deps`, `ruff check ppvm-python/`

## Notes

- Closes #94: the `u1-conserving` branch and `exchange` / `xyzz` gates are no longer needed. The deferred-truncate idiom in this PR replaces them; see `test_truncate_false_pair_reproduces_old_exchange` for the one-line migration.
- The `u1-conserving` branch will be deleted from the remote once this PR lands.

### DIFF
--- a/crates/ppvm-python-native/src/interface_tableau.rs
+++ b/crates/ppvm-python-native/src/interface_tableau.rs
@@ -45,57 +45,91 @@ macro_rules! create_interface {
                 self.inner.measure(addr0)
             }
 
+            // All gate methods take a `truncate: bool = true` kwarg
+            // purely for API symmetry with `PauliSum` (which uses it to
+            // defer the auto-truncate). `GeneralizedTableau` has no
+            // analogous truncation step — the tableau representation is
+            // exact — so the kwarg is silently ignored here. Keeping the
+            // signature parallel lets the Python `RotationsMixin` /
+            // `CliffordMixin` / etc. work uniformly across both backends.
+
             // clifford
-            pub fn x(&mut self, addr0: usize) {
+            #[pyo3(signature = (addr0, truncate = true))]
+            pub fn x(&mut self, addr0: usize, truncate: bool) {
+                let _ = truncate;
                 self.inner.x(addr0);
             }
 
-            pub fn y(&mut self, addr0: usize) {
+            #[pyo3(signature = (addr0, truncate = true))]
+            pub fn y(&mut self, addr0: usize, truncate: bool) {
+                let _ = truncate;
                 self.inner.y(addr0);
             }
 
-            pub fn z(&mut self, addr0: usize) {
+            #[pyo3(signature = (addr0, truncate = true))]
+            pub fn z(&mut self, addr0: usize, truncate: bool) {
+                let _ = truncate;
                 self.inner.z(addr0);
             }
 
-            pub fn h(&mut self, addr0: usize) {
+            #[pyo3(signature = (addr0, truncate = true))]
+            pub fn h(&mut self, addr0: usize, truncate: bool) {
+                let _ = truncate;
                 self.inner.h(addr0);
             }
 
-            pub fn s(&mut self, addr0: usize) {
+            #[pyo3(signature = (addr0, truncate = true))]
+            pub fn s(&mut self, addr0: usize, truncate: bool) {
+                let _ = truncate;
                 self.inner.s(addr0);
             }
 
-            pub fn s_adj(&mut self, addr0: usize) {
+            #[pyo3(signature = (addr0, truncate = true))]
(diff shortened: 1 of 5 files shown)


crates/ppvm-python-native/src/interface_tableau.rs
```diff
@@ -45,57 +45,91 @@ macro_rules! create_interface {
                 self.inner.measure(addr0)
             }
 
+            // All gate methods take a `truncate: bool = true` kwarg
```
Roger-luo, line 48: technically we can have a truncate step tho? @david-pl

Roger-luo (approved):
I think this is an easy PR, I think it would be nice to add a truncate option for tableau too just for completeness before the final release. @david-pl

crates/ppvm-python-native/src/interface_tableau.rs
david-pl: Yes, both options are valid. Just have to decide. I still think it's nice to have the default behavior be to automatically truncate after every gate. When working with this, I noticed that this is usually what you want since it gives you the best performance. That's why I bundled it initially. It's also how PP.jl does it.

If we make it a separate step, the user would have to add that after every gate to recover the current default behavior. Is there a more convenient way to have the truncate after each gate behavior when it's a separate step? Otherwise, I'm leaning towards the kwarg approach here.

david-pl (approved):

PR #109, opened 2026-06-08
Split up mixins so we separate truncate kwarg

david-pl:
This slipped through my review of #100. That PR added a dead truncate kwarg on the `GeneralizedTableau`, which was part of the public method, but didn't do anything. This would be quite surprising to users.

Also, this currently breaks CI for #101.


Roger-luo (approved):
oh nvm this is an easy PR to review

PR #117, opened 2026-06-12
perf(runtime): faster map_insert + specialized diagonal 2-qubit rotations

Roger-luo:
## Summary

Performance pass on the `PauliSum` propagation hot path (as exercised by the
Trotter benchmark), reduced to the changes that are *correct under all usage*.

### Kept optimizations
1. **`rehash` byte-slice hashing** (`word/data.rs`) — feed the raw
   `xbits`/`zbits` storage bytes through the hasher's byte-slice path so
   FxHasher consumes 8 bytes per round instead of one byte at a time. Made
   sound for arbitrary storage via a `bytemuck::Pod` bound on `PauliStorage`
   and safe `bytemuck::bytes_of` (no `unsafe`) — see the soundness note below.
2. **Specialized `rzz` / `rxx` / `ryy` fast paths** (`sum/rot2.rs`) —
   hand-written diagonal two-qubit rotations that bypass the generic `comm_2`
   commutator: a single x/z-bit comparison detects commuting terms and
   short-circuits, flipping only the relevant bits. Loss fallbacks match the
   generic `rotate_2` path exactly (dead-code-eliminated for non-lossy words).
   Verified bit-exact against `rotate_2` for every two-qubit input Pauli.
3. **Scratch-`Vec` `map_insert`** (`sum/data.rs`) — collect produced terms into
   a reusable `Vec` (cheap pushes, no hashing) and merge once, dropping a
   hashmap probe + insert per produced term and making the per-call clear O(1).
4. **DashMap `map_insert_vec` via `ParallelExtend`** (`map/dashmap.rs`) — extend
   the destination `Vec` directly from the parallel iterator, avoiding an
   intermediate `collect` allocation (Copilot autofix).

### Reverted: `Strategy::discard` insertion-pruning
An earlier iteration skipped inserting gate-produced terms a *following*
`truncate` would drop, to avoid IndexMap's O(T) hash-index rebuild. As the
Copilot review flagged, this is **only** a no-op when `truncate()` is the
immediate next call. With truncation deferred across gates, two sub-threshold
contributions to the same key can sum above the threshold (`|a + b| ≥ τ` while
`|a|, |b| < τ`), so dropping either at production loses a genuinely surviving
term and biases the terms it would have fed — the same contract violation the
`contains_with` revert (1df64df0) addressed within a single merge, but spread
across gates. It also lacked a test. It has been **removed**: `map_insert`
again merges every produced term unconditionally, and truncation is the only
place terms are dropped. This was the dominant share of the earlier headline
numbers, so the figures below are substantially smaller — but correct.

### Soundness note (`rehash`)
The old fast path reinterpreted storage `A` as `&[u8]` via
`slice::from_raw_parts`, which is UB if `A` carries padding. The blanket
`PauliStorage` impl didn't preclude that. Adding `bytemuck::Pod` to the bound
makes the byte view safe by construction; every concrete storage
(`[u8; N]`, `[u64; N]`) already qualifies, and the bytes hashed are identical
(hash values, iteration order, and the `pauli_sum_creation-5` snapshot are
unchanged).

## Benchmarks

Measured with criterion against `main` (`417980cb`); bench drivers are
byte-identical to `main`.

### `trotter` — 12 qubits, all 6 hashmap configs
*criterion defaults (100 samples, 5 s); median times*

| Config | main | this branch | change |
|---|---|---|---|
| FxIndexMap<2> | 341.0 µs | 298.4 µs | −12.2% |
| GxIndexMap<2> | 523.4 µs | 332.3 µs | −36.1% |
| FxHashMap<2> | 376.3 µs | 355.0 µs | −4.3% |
| GxHashMap<2> | 361.7 µs | 358.8 µs | −1.3% |
| FxDashMap<2> | 9.458 ms | 7.797 ms | −17.6% |
| GxDashMap<2> | 8.740 ms | 8.260 ms | −5.5% |

### `trotter-scaling` — FxIndexMap<8>, qubit sweep
*10 samples, 3 s measurement, identical settings both sides; % change vs main*

| Qubits | 2 | 6 | 10 | 14 | 18 | 22 | 26 | 30 | 34 | 38 | 42 | 46 | 50 | 54 | 58 | 62 |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| Δ | −29.8% | −14.5% | −15.4% | −14.8% | −15.7% | −11.8% | −8.7% | −9.9% | −9.4% | −8.1% | −10.7% | −6.2% | −3.3% | −1.6% | −1.1% | −2.0% |

The win is ~12–16% at small/mid term counts (where the per-term constants —
`rehash`, `rzz`, scratch buffer — dominate) and tapers to ~1–2% at 62 qubits
(where `truncate`'s O(T) rebuild, which the reverted pruning used to skip,
dominates again). All changes are significant (p < 0.05).

## Testing
- `tests/trotter.rs` — end-to-end golden for the exact expectation value
  (gate-math regression) plus a `CoefficientThreshold(1e-6)` truncation-fidelity
  bound.
- `tests/truncation_semantics.rs` (new) — deferred-truncation guard: two `rx`
  on a single `Z` with no truncate between must keep the merged above-threshold
  `Y` term, matching a non-truncating reference. This is the regression test the
  reverted pruning lacked; it fails with that optimization present.
- `rot2` unit tests — each fast path matches the generic `rotate_2` for every
  two-qubit input Pauli, plus explicit hand-computed `rzz` values and a
  non-adjacent-addressing case.


Roger-luo:
autotune shows hash can be improved and rzz/rxx can be further specialized, I manually reverted 2 of the changes due to missing proper test cases and we are having an e2e test for trotter

Roger-luo:
bump @david-pl can I use a second eye here? I added a new test due to missing harness guarding observable values, autotune tried to discard small amplitudes at insertion time, which I had to manually revert that proposal.

crates/ppvm-runtime/src/traits/storage.rs
```diff
@@ -32,7 +39,8 @@ impl<
         + Sync
         + std::fmt::Debug
         + Ord
-        + PartialOrd,
+        + PartialOrd
+        + bytemuck::Pod,
```
Roger-luo, line 43: FYI. the key of this optimization is assuming padding for the pauli string encoding, which I assume we may want similar optimization for other things. 

cc: @david-pl

david-pl: Definitely something we want for `GeneralizedTableauSum`, where we hash the entire tableau.

david-pl (approved):
LGTM

PR #129, opened 2026-06-17
perf(ppvm-python-native): use mimalloc as the global allocator

AlexSchuckert:
## What

Sets [`mimalloc`](https://github.com/microsoft/mimalloc) as the `#[global_allocator]` for the `ppvm_python_native` extension module.

## Why

mimalloc returns freed pages to the OS more aggressively than the default system allocator. This materially reduces **peak RSS** on the allocation-heavy Pauli-propagation paths, where each gate / truncation step churns large transient `Vec` / `HashMap` buffers that the system allocator tends to retain.

Split out of #98 per review feedback (@Roger-luo): the allocator is an independent optimization and shouldn't be bundled into a feature PR — this lets it be reviewed and benchmarked on its own.

## Changes
- `crates/ppvm-python-native/Cargo.toml`: add `mimalloc = { version = "0.1", default-features = false }`.
- `crates/ppvm-python-native/src/lib.rs`: declare the global allocator (with a justifying comment).

## Validation
- `maturin develop --release` builds and links cleanly.
- `pytest ppvm-python/test/test_basics.py` green (smoke).

No API or behavior change — allocator only.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -645,6 +645,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "58f929b4d672ea937a23a1ab494143d968337a5f47e56d0815df1e0890ddf174"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.49"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a45a52f43e1c16f667ccfe4dd8c85b7f7c204fd5e3bf46c5b0db9a5c3c0b8e9"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "linux-raw-sys"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Roger-luo:
Ran an allocator A/B for this PR locally on Linux x86_64.

Setup:
- Compared PR base `ea75072a` against PR head `4dd7925`.
- CPython 3.12.3.
- Release build of `ppvm-python-native` with `RUSTFLAGS='-C target-feature=+aes,+sse2 -L native=/usr/lib/python3.12/config-3.12-x86_64-linux-gnu'`.
- Imported the built `ppvm_python_native` extension directly and exercised the native `PauliSumIndexMapFxHash*` classes. The PR `.so` contains mimalloc strings; the baseline `.so` does not, so the comparison did exercise the allocator change.

Results:

| Workload | Baseline | PR/mimalloc | Takeaway |
| --- | ---: | ---: | --- |
| 12q Ising/Trotter, 200 rounds, 10 warmups, `hyperfine` mean | `1.9168s ± 0.0529` | `1.9165s ± 0.0637` | no meaningful throughput difference |
| Same workload, RSS | ~19.6 MiB | ~23.4 MiB | mimalloc higher |
| 20q ladder stress, 15 steps | `51.40s`, HWM ~112.7 MiB, final RSS ~19.5 MiB | `49.61s`, HWM ~172.8 MiB, final RSS ~163.0 MiB | small speedup, much worse RSS |
| 18q ladder stress, 15 steps | `50.87s`, HWM ~112.8 MiB, final RSS ~19.5 MiB | `49.62s`, HWM ~172.7 MiB, final RSS ~162.8 MiB | same memory pattern |

Conclusion: these Linux results do not support the stated RSS/memory-throughput rationale. Throughput is neutral on the small Trotter workload, and the memory-stress workload retains substantially more resident memory with mimalloc. Closing this PR rather than taking the allocator change as-is.

PR #116, opened 2026-06-12
Batch versions of all clifford gates

david-pl:
Closes #99.

Makes batched versions proper traits and completes the implemented set. Also uses batched gates for STIM.

PR #130, opened 2026-06-17
feat: mirror stim.TableauSimulator API (broadcasting, renames, measurement record)

rafaelha:
## Summary
Reshapes ppvm's gate/measurement/noise API to mirror `stim.TableauSimulator`, across both backends (Pauli-propagation `PauliSum` and the `Tableau`/`GeneralizedTableau` stabilizer backends), the PyO3 bindings, and the Python package.

- **Broadcasting** via a new `Targets` trait — `h(0)` or `h([0,1,2])`; two-qubit gates broadcast over consecutive pairs (`cnot([0,1,2,3])` in Rust; `*targets` varargs in Python).
- **Renames:** `s_adj`→`s_dag`, `sqrt_x_adj`→`sqrt_x_dag`, `sqrt_y_adj`→`sqrt_y_dag`, `t_adj`→`t_dag`, `depolarize`→`depolarize1`.
- **New surface:** `x_error`/`y_error`/`z_error`, stim aliases `cx`/`zcx`/`zcz`/`zcy`, `measure_many`, `current_measurement_record`, `reset_x`/`reset_y`/`reset_z`.
- `measure` stays single-target and loss-aware (`LOST` is a real outcome, not collapsed to `bool`).
- Python: `*targets` varargs, keyword-only `theta=`/`p=`, `truncate=` preserved for `PauliSum`.
- **Deliberately not mirrored** (semantic mismatch with a coefficient-weighted / fixed-size state): `peek_*`, `current_inverse_tableau`, `canonical_stabilizers`, `state_vector`, `measure_kickback`, `set_state_from_*`, auto-growing `num_qubits`.

### Cross-cutting fixes caught during review
- Measurement record stays **neutral** across `reset` and loss channels (they pop the entry their internal `measure` pushes).
- Batch gate fast-path falls back to per-qubit application for storage wider than 8 words (fixes a panic on `U256`-backed tableaux).
- `MR` records the **reported** (post-readout-noise) value, consistent with `M`.

Design & plan: `docs/superpowers/specs/2026-06-17-stim-mirror-api-design.md`.

## Test Plan
- [x] `cargo test --workspace` — 620 passed, 0 failed
- [x] `cargo test --workspace --doc` — 17 passed
- [x] `cargo clippy --workspace -- -D warnings` + `cargo fmt --check` — clean
- [x] Python `pytest` — 127 passed; `ruff format`/`ruff check` — clean
- [ ] `prek run --all-files` in CI (local prek 0.3.0 predates the repo's `prek.toml` format)

## Known follow-ups (non-blocking)
- Rust two-qubit gates accept `cnot([2,3])` only, not `cnot(2,3)` (Rust has no overloading/variadics); Python supports `cnot(2,3)`, `cnot(2,3,4,5)`.
- A dedicated broadcast-equivalence test and a wide-storage batch-equivalence test would tighten coverage.


david-pl:
@rafaelha I like the idea of the `Targets` trait in principle. But be aware that there is another PR #116, which adds dedicated versions of Clifford gates that are broadcasted over multiple qubits, which leads to significant performance gain. As far as I can tell, in this PR you just loop over lists of targets. Also, there is the somewhat orphaned impl for a few batched Clifford gates which are used for the msd-fused benchmark specifically. The referenced PR makes this a proper trait, whereas here this is left unchanged I think.

Do you see a good way to reconcile these approaches?

Roger-luo:
Superseded by #134, which ports this Stim API mirroring work onto the current crate layout after the `ppvm-runtime` split.

The new branch keeps Rafael credited as co-author in the commit trailer.

Roger-luo:
I think the broadcast API in Python might make sense for the sake of STIM as a de facto standard, but it is not really idiomatic to Rust. I would say that Rust users prefer to use an explicit name or the builder pattern instead of implicitly inserting specialization methods.

Essentially, the behavior of the function is semantically different when you are accepting a vector versus a number, so it does not make sense to keep the same name for the method.

rafaelha:
Thanks Roger. Closing this in favor of #134

PR #132, opened 2026-06-19
Mix hasher to avoid overly large buckets

david-pl:
While running benchmarks, there was an odd increase of runtime when the storage chunks were largely filled. Claude figured out that this was because the hash bits are highly correlated then, leading to overly large buckets.

Before:

image

After:

image

**Note**: the fix here is simple, but it actually regresses performance for _some_ storage types. However, the "hot path", which is the one used by the python bindings, clearly benefits.


Roger-luo:
I think we should create proper abstractions instead of prioritizing or specializing for a hot path. I am okay with not favoring less useful use cases, but in order to grow the framework healthily, we should not regress on other cases. I think we should create an abstraction here to help with specialization.

Roger-luo:
## gxhash vs fxhash on the TFIM Trotter scaling (independent reproduction)

Following up on the storage-cliff discussion — here is the same TFIM Trotter
qubit-count sweep, now with **gxhash** alongside fxhash *with* and *without*
the hash fold, plus **PauliPropagation.jl** as a reference line.

tfim_trotter_scaling


**Setup** (all four series, identical circuit): TFIM, J=1.0, h=1, dt=0.1, 20
Trotter steps, truncation 1e-6, depolarizing noise 1e-4. Storage tier is
dispatched by qubit count exactly like the Python bindings
(`[u8;2]`→`[u8;4]`→`[u8;8]`→`[u8;16]`→`[u8;32]`). Term counts match ppvm and
PP.jl exactly at every point. Both ppvm and PP.jl run single-threaded.

| qubits | tier | fxhash (no fold, pre-PR) | fxhash + fold (this PR) | gxhash | PauliPropagation.jl |
|-------:|:----:|------:|------:|------:|------:|
| 32 | `[u8;8]`  | 0.21 s | 0.08 s | 0.08 s | 0.90 s |
| 48 | `[u8;8]`  | 1.03 s | 0.18 s | 0.18 s | 3.16 s |
| 56 | `[u8;8]`  | **1.79 s** | 0.26 s | 0.24 s | 4.51 s |
| 60 | `[u8;16]` | 0.32 s | 0.32 s | 0.30 s | 5.36 s |
| 122 | `[u8;32]`| 1.50 s | 1.50 s | 1.50 s | 31.5 s |

**Takeaways**
- **fxhash, no fold (pre-PR)** reproduces the bump: runtime climbs through the
  `[u8;8]` tier, peaks at **56 qubits (~1.8 s)**, then drops at 60 qubits when
  dispatch moves to `[u8;16]`.
- **gxhash** and **fxhash + fold (this PR)** are both smooth — no bump. gxhash
  avalanches well even on short keys (AES), so it never clustered and never
  needed the fold; folded fxhash tracks it.
- ppvm (either fixed hasher) is ~10–20× faster than PauliPropagation.jl in this
  configuration.

Harness committed under `benchmarks/` (see `benchmarks/README.md`).

> ⚠️ **These numbers are _not_ the same as the original benchmark in this PR.**
> They were produced independently on different hardware, with high-fill
> parameters (J=1.0, 20 steps) chosen specifically to reach the regime where
> the cliff appears — the J=1/8 mirror params keep the state at ~1e3 terms, too
> small to bump — and with PauliPropagation.jl pinned to a single thread. Read
> them as a qualitative reproduction of the *shape*, not a re-run of the
> original measurements.

<sub>Analysis and plot produced by Claude (Claude Code), under @Roger-luo's instruction.</sub>

Roger-luo:
I did a bit of a change in terms of hasher abstraction. The behavior you see in the original benchmark is mainly because of how FXHash is engineered (specifically the rustc implementation). 

There is a different one called GxHash, which already uniformly distributes the bit bucket. Without any tweaks, GxHash already gives perfect results. 

Essentially, this PR specialization is more for those hashers that do not uniformly distribute the workload.

@david-pl I'm gonna merge this PR now I think it should be good to go, we can revisit after open-source if anything we want to modify.

david-pl:
@Roger-luo The abstraction here is fine, so I don't think we need to change anything. Just to note, I intentionally did not go for the abstraction in the first place, because we don't really use the other hashers that much, so I preferred the simple solution. The abstraction you added also adds some complexity that I'm not sure will be worth it in the end. Let's see how this goes in the long term though.

PR #134, opened 2026-06-22
[codex] feat(stim): mirror stim simulator api after crate split

Roger-luo:
Supersedes #130. Ports Rafael's Stim API mirroring onto the post-`ppvm-runtime` layout, settles the target-passing convention, and consolidates the Python packaging.

### Highlights
- **Rust gate API** — scalar methods + explicit `*_many` broadcast helpers (`x_many`, `cnot_many`, `measure_many`), mirroring Stim's `_many` naming while keeping scalar-vs-collection dispatch explicit. (Renamed from `*_batch`; the now single-use `Targets` trait is gone.)
- **Python gate API** — targets accept variadic ints **or** a single sequence/ndarray, merging Stim's `TableauSimulator` (`x(0, 1, 2)`) and `Circuit.append` (`x([0, 1, 2])`) conventions. `measure` stays scalar; `measure_many` broadcasts.
- **Packaging** — one `ppvm` wheel (maturin mixed layout); the compiled extension is the private `ppvm._core`, so users only `import ppvm`. Shipped as a single `cp310-abi3` wheel per platform. Plain `cargo build` also links the cdylib (macOS `build.rs`).
- **CI** — a matrix now verifies build/link/tests on **Linux, macOS, and Windows**.

<details>
<summary>Python target-passing examples</summary>

```python
# single-qubit gates (x, y, z, h, s, sqrt_x, t, ...)
t.x(0)                       # single int
t.x(0, 1, 2)                 # variadic ints
t.x([0, 1, 2])               # list / tuple / range
t.x(np.array([0, 1, 2]))     # ndarray

# two-qubit gates (cnot, cz, cy) — flat, consumed as consecutive pairs (even count)
t.cnot(0, 1)                 # CX(0,1)
t.cnot(0, 1, 2, 3)           # CX(0,1), CX(2,3)
t.cnot([0, 1, 2, 3])         # same, as a sequence
t.cnot(np.array([0, 1, 2, 3]))

# rotations (rx/ry/rz, rxx/ryy/rzz) + theta (positional or keyword)
t.rx(0, 1, 2, theta=0.1)
t.rx([0, 1, 2], 0.1)
t.rxx([0, 1, 2, 3], theta=0.1)    # pairs + theta

# noise (x_error, pauli_error, depolarize2, ...) + p
t.x_error([0, 1, 2], p=0.01)
t.pauli_error(np.array([0, 1]), p=[0.01, 0.0, 0.02])

# measurement — measure stays scalar; measure_many broadcasts
t.measure(0)                 # -> single result
t.measure_many(0, 1, 2)      # -> list
t.measure_many([0, 1, 2])    # or a sequence / ndarray
```

Rust callers mirror these via `*_many(&[usize])` / `*_many(&[(usize, usize)])`.
</details>

<details>
<summary>Packaging &amp; build details</summary>

- `ppvm-python/pyproject.toml` → `build-backend = "maturin"` with `[tool.maturin]` `manifest-path` → the crate, `python-source = "src"`, `module-name = "ppvm._core"`. The standalone `ppvm-python-native` distribution and top-level `ppvm_python_native` module are removed.
- `pyo3/abi3-py310`: one stable-ABI wheel loads on CPython 3.10+ instead of one per Python version.
- `ppvm-python-native/build.rs` adds the macOS `-undefined dynamic_lookup` flag, scoped to the cdylib, so `cargo build --workspace` links the extension without going through maturin (maturin still builds the wheels and sets the same flag).
- The native crate stays a Cargo workspace member; the developer guide and `AGENTS.md` are updated to describe the single mixed wheel.
- Also: keeps Stim `MPAD` results and the tableau measurement record in sync, and makes workspace build/test env-free (gxhash target features live arch-scoped in `.cargo/config.toml`).
</details>

Rafael is credited in the commits: `Co-authored-by: Rafael Haenel <rhaenel@quera.com>`.

**Validation:** `cargo build --workspace`, `cargo test --workspace`, `cargo check --workspace --all-targets`, `uv run --project ppvm-python --group dev pytest` (167 passed), and the full pre-commit suite (rustfmt, clippy, ruff, ty, hawkeye) — all green. CI now also exercises macOS and Windows.


Roger-luo:
Updated this PR for the Rust API design feedback and moved the scaffold-spec rationale into the PR body:

- Rust gate, rotation, reset, and noise methods are scalar again (`cnot(0, 1)`, `rzz(0, 1, theta)`, etc.).
- Python keeps the Stim-style broadcast surface; PyO3 wrappers now dispatch target lists through explicit Rust `*_batch` helpers.
- Rust call sites, examples, docs, and skill snippets were updated to use scalar methods or explicit batch helpers.
- The superpower scaffold spec is intentionally not included in the PR diff; the PR body now summarizes the design and rationale instead.

Validation:
- `RUSTFLAGS="-C target-feature=+aes,+sse2" cargo check --workspace --all-targets`
- `RUSTFLAGS="-C target-feature=+aes,+sse2" cargo test --workspace --exclude ppvm-python-native`
- `UV_CACHE_DIR=/tmp/uv-cache RUSTFLAGS="-C target-feature=+aes,+sse2" uv run --project ppvm-python --group dev pytest ppvm-python/test -q`

`cargo test --workspace` still only fails at the local PyO3 test link step because this machine is missing `libpython3.12`. `npm run build` under `docs/` is blocked locally because `npm` is not installed.

david-pl (commented):
_Findings by Claude_

## Code review

Thorough, mostly-mechanical migration with one genuinely thoughtful new feature (the measurement record). The rename + broadcast migration is **complete and internally consistent** — I grepped the whole tree and found no leftover scalar callers of renamed methods in functional code (the `squin.s_adj` / `squin.depolarize` hits are the squin DSL frontend, correctly translated by the interpreter to `_dag` / `depolarize1`). `MeasurementResult(int)` mapping (0/1/2) lines up exactly with `measurement_to_u8`, and backward compat holds where it counts (`tab.x(0)`, `tab.cnot(0, 1)` still work).

One issue I'd want resolved or consciously documented before merge (#1); the rest are quality/consistency nits.

### 1. `MPad` diverges from the measurement record (medium — correctness of the new feature)
`crates/ppvm-stim/src/executor.rs:632`

```rust
ExtendedInstruction::MPad { bits, prob, .. } => {
    for &bit in bits {
        results.push(Some(tab.flip_with_prob(bit, noise)));  // → results, but NOT measurement_record
    }
}
```

Every other measurement path (`M`/`MZ` via `measure_noisy`, `MR` via the explicit overwrite) keeps `results` and `tab.measurement_record` in lockstep — that's the invariant the new test asserts ("the executor's returned results must equal the tableau's record exactly"). `MPad` is the exception: its bits enter `results` but never the record, so after any program containing `MPAD`, `current_measurement_record() != results`. Stim's `MPAD` *does* append to the record (loss, by contrast, is correctly excluded from **both**), so this looks like a real gap rather than an intentional exclusion, and it's untested. Either push these bits onto `measurement_record` too, or document the exclusion and relax the test's "exactly" claim.

### 2. Breaking change headline + odd-target error UX
`theta` and `p` are now **keyword-only** in Python (`rx(*targets, theta)`, `depolarize1(*targets, p)`), so `ps.rx(0, 0.1)` and `ps.depolarize(0, 1e-3)` both break. Worth a "Breaking changes" section in the PR description / changelog. Relatedly, an odd target count to a two-qubit gate panics through Rust (`flat_pairs`'s `assert!`, `Targets::pairs`'s `.expect`) and surfaces as `pyo3_runtime.PanicException` rather than a `ValueError` — a mapped exception would be friendlier for ordinary user error.

### 3. Redundant manual impls in `GeneralizedTableau` Clifford (low — DRY)
`crates/ppvm-tableau/src/gates/clifford.rs:175-225`: `h`, `cz`, `sqrt_x`, `sqrt_x_dag`, `sqrt_y`, `sqrt_y_dag` are hand-expanded with bodies *identical* to what `impl_generalized_tableau_clifford!` / `..._pair!` produce, while their siblings (`x`, `y`, `z`, `s`, `cnot`, `s_dag`, `cy`) use the macros. This PR actually converted them *from* macro calls *to* manual code (~30 lines of duplication) for no functional reason — folding them back into the macros would be more consistent.

### 4. Naming inconsistency: `*_dag` scalar vs `*_adj_batch` (low)
Scalar daggers were renamed (`s_dag`, `sqrt_x_dag`, …) but the batch counterparts kept the old stem (`s_adj_batch`, `sqrt_x_adj_batch`, `sqrt_y_adj_batch`). Internal-only, so a nit — but since this is the PR introducing the `_dag` convention, renaming the batch stems here avoids leaving a permanent inconsistency.

### 5. The `Targets` trait is nearly dead code (low — simplification)
`crates/ppvm-traits/src/traits/targets.rs` adds a full `Targets` trait with `each()` + `pairs()`, but the only consumer anywhere is `measure_many` (via `each()`); `pairs()` has no non-test caller, and every gate site uses `*_batch(&[usize])` + the separate `flat_pairs` helper instead (which duplicates `Targets::pairs`). Either route the batch methods / `flat_pairs` through `Targets`, or drop `Targets` and give `measure_many` a plain `&[usize]`.

### 6. Minor
- **Uneven broadcasting**: `amplitude_damping`, `loss_channel`, `correlated_loss_channel`, `reset_loss_channel` stay scalar while everything else broadcasts. Defensible (non-Stim ops), but worth a one-line note so it reads as deliberate.
- **Stale generated docs**: `docs/src/data/python-api.json` and `docs/src/generated/notebooks/*.html` still show `t_adj` / `sqrt_y_adj` / old signatures. Regenerated artifacts (no `npm` in CI container), so follow-up not blocker — but the published API ref is wrong until regenerated.
- **`SKILL.md` tables** cover the renames but don't list the new aliases (`cx`/`zcx`/`zcz`/`zcy`), `reset_x/y/z`, `measure_many`, `current_measurement_record`, or the `do = run` alias.

### Test coverage
Good for renames, broadcast, keyword-only params, and the basic record. Gaps: no MPad-record test (see #1); `test_renamed_methods_exist` only checks `hasattr`, not multi-target state correctness; no test that an odd target count raises a clean Python error.

Nice work overall — the measurement-record neutrality handling for reset/loss and the readout-noise overwrite are clean and well-tested.

Roger-luo:
~@rafaelha do you know what's the reason STIM has `measure_many` instead of broadcasting `measure` in the first place?~

this is because

```
measure -> int
measure_many -> list[int]
```

if defined as broadcast by default the typing rule will be ugly.

for context, I'm trying to clean up the rust APIs - we prob don't have to match STIM in rust, match it in Python is ok. I don't want to make dispatch implicit in rust it's not very idiomatic and could cause confusion when we optimize the behaviour in the future.

I think maybe the cleaner solution should be rename `*_batch` to `*_many` instead so we both match STIM, and being explicit and idiomatic about the behaviour in rust

rafaelha:
Yes exactly. In a TableauSimulator, one may often want to do `if sim.measure(7): ...`.

Sure, I like renaming `_batch` to `_many` in Rust

Roger-luo:
another issue I see in STIM convention is the vararg, in STIM you have

```
t.h(0, 2)
```

representing broadcasting, but if you have a numpy array, you have to write

```
t.h(*np_arr)
```

which breaks the numpy broadcast convention (the function acting directly on the numpy array should broadcast), we can simply solve this by also accepting a `Sequence` object in terms of single argument input

Roger-luo:
I also include some clean-ups on the python extension setup, don't want to split them in a separate PR given they are simple. will merge once CI passes

PR #135, opened 2026-06-22
Add measure_batch and adopt it for STIM measurement

david-pl:
Adds `measure_batch` to `ppvm-tableau` — measure a chosen set of qubit indices,
the analogue of the batched Clifford gates from #116. `measure_all` becomes the
`0..n_qubits` case (behavior unchanged, no regression), and the noiseless `M`/`MZ`
path in `ppvm-stim` now uses it, reusing the measurement scratch across targets
the same way #116 did for gates. Noisy readout and `MR` keep the per-qubit loop,
so RNG draws are unchanged.

Also adds a samply profiling harness (`scripts/profile_measure_batch.sh`) and a short findings doc. The cost splits into two regimes: a constant
~16µs O(n²) tableau floor (`compute_decomposition` + `update_tableau`), and the
case-a `Vec`↔`HashMap` round-trip (~35% at ~1000 coefficients). No optimization
applied yet — candidates are written up for follow-up.

**Note**: Rust-only — `measure_all` isn't exposed to Python, so no binding changes.

FYI, @rafaelha @Roger-luo we may want to consider the Python bindings for this in #134.


scripts/samply_top.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env python3
```
Roger-luo, line 1: What does this file do? If it renders something, maybe we can post the render result in this PR just for future reference.

Roger-luo:
Superseded by #142, which carries this work forward onto the `measure_many` API that #134 established (folding the scratch-reusing implementation in as the tableau's `LossyMeasure::measure_many` override rather than a parallel `measure_batch`). Your commits are preserved there and you're credited as co-author. Closing in favor of #142.

PR #138, opened 2026-06-22
fix: gate 64-bit-only storage and threaded parser for wasm32

rafaelha:
## Summary

Makes the affected crates compile for `wasm32` (32-bit pointer width, no OS threads) without changing behavior for 64-bit native builds.

- **`ppvm-pauli-sum`**: gate the `fx64hash` configs behind `target_pointer_width = "64"`. They use `[u64; N]` Pauli storage, which `bitvec` only supports on 64-bit targets.
- **`ppvm-pauli-word`**: gate the `u64`-hardcoded `From<&str>`/`From<String>` `PhasedPauliWord` constructors (and the `num::Integer` import they need) behind `target_pointer_width = "64"`. Generic-`A` code paths are unaffected on every target.
- **`stim-parser`**: `wasm32` has no OS threads, so run the parser inline there instead of on a dedicated 16 MiB stack thread (which relies on `std::thread`).

## Impact

These are pure `#[cfg]` additions. On 64-bit native targets every gate evaluates to "include the original code," so compilation and behavior are unchanged. Only 32-bit targets observe a difference, and those did not compile before.


Roger-luo:
superceded by #143 the crates still require further changes to be wasm compatible, I'm gonna address this in a separate pr

PR #144, opened 2026-06-23
feat(wasm): build the full workspace (minus python-native) for browser wasm32

Roger-luo:
## Summary

Follow-up to #143 (now merged): #143 made the *compile gates* generic; this PR
closes the **dependency / feature-graph** blockers so the whole workspace
except `ppvm-python-native` cross-compiles to **`wasm32-unknown-unknown`**
(browser / wasm-bindgen) — `cargo build --target wasm32-unknown-unknown` with
**no extra flags**. Native builds keep gxhash/dashmap/ahash/rayon and are
byte-for-byte unchanged.

## Mechanism (automatic)

Cargo prunes a dependency under `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]`
on wasm, but the **feature named after it stays on**, so `#[cfg(feature = "X")]`
code would still try to compile against the absent crate. The fix is two
coordinated moves per native-only dep: target-gate the dep **and** gate its code
sites with the same `not(target_arch = "wasm32")`. The features then go inert on
wasm, so **no `default-features = false` propagation through the inter-crate deps
is needed**.

## Changes

- **`ppvm-traits`** — target-gate `ahash`/`dashmap`/`gxhash`/`rayon`; gate the
  dashmap map module, the ahash `ACMap` impls, and the `gxhash` `HashFinalize` impl.
- **`ppvm-pauli-word`** — target-gate `gxhash`.
- **`ppvm-pauli-sum`** — target-gate `dashmap`/`gxhash`; gate the
  `config::dashmap` module and `config::indexmap`'s `ByteGxHash`/`ByteGxHashF64`
  (also fixes a latent `indexmap`-without-`gxhash` coupling).
- **`ppvm-tableau-sum`** — target-gate `gxhash`/`rayon`; `word_fingerprint`
  falls back from `GxHasher` to `fxhash::FxHasher` on wasm (a transient in-memory
  dedup hash, resolved by `structurally_equal`, never persisted or compared
  across platforms).
- **`ppvm-tableau` / `ppvm-stim`** — target-gate `rayon` (no OS threads on wasm).
- **rand entropy** — `rand::make_rng()` has no default source on
  `wasm32-unknown-unknown`, so select getrandom's `wasm_js` (Web Crypto) backend
  via a `--cfg getrandom_backend="wasm_js"` rustflag in `.cargo/config.toml` plus
  the `wasm_js` feature in `ppvm-tableau`'s wasm-only dep table; feature
  unification covers tableau-sum / stim / top-level `ppvm`.
- **CI** — new `wasm32 build (browser)` job cross-compiles the workspace on every PR.
- **docs** — WebAssembly subsection in `develop.astro`.

## Verification

- Native `cargo test --workspace` (excl. python-native's macOS test-link issue):
  **744 passed, 0 failed**.
- Native `--all-targets` and the `ppvm-python-native` cdylib build unchanged.
- Full workspace builds for `wasm32-unknown-unknown` — incl. `ppvm-tableau`,
  `ppvm-tableau-sum`, `ppvm-stim`, and the top-level `ppvm` crate.
- On wasm the dependency tree drops `ahash`/`gxhash`/`getrandom 0.3` entirely,
  leaving only `getrandom 0.4` (`wasm_js`) for `rand`.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,3 +2,10 @@
 # build/test commands match CI without per-command RUSTFLAGS.
 [target.'cfg(target_arch = "x86_64")']
 rustflags = ["-C", "target-feature=+aes,+sse2"]
+
+# On wasm32 (browser) `getrandom` has no default entropy source. Select its JS
+# backend so `rand` (e.g. `rand::make_rng()` in the tableau simulators) reads
+# from the Web Crypto API. Pairs with the `getrandom`/`wasm_js` feature enabled
+# in ppvm-tableau's wasm-only dependency table. No per-command RUSTFLAGS needed.
+[target.wasm32-unknown-unknown]
+rustflags = ["--cfg", 'getrandom_backend="wasm_js"']
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,28 @@ jobs:
       - name: Run ppvm-stim tests with rayon feature
         run: cargo test -p ppvm-stim --features rayon
 
+  # Cross-compile the whole workspace for browser wasm so a wasm regression
+  # surfaces in CI. `ppvm-python-native` is a CPython extension (never a wasm
+  # target) and is excluded. Native-only acceleration deps (gxhash, dashmap →
+  # rayon, ahash) are pruned automatically by the `cfg(not(target_arch =
+  # "wasm32"))` dependency tables, and `rand`'s entropy uses the getrandom
+  # `wasm_js` backend wired in `.cargo/config.toml` — so no extra flags here.
+  wasm-build:
+    name: wasm32 build (browser)
+    needs: pre-commit
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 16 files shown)


Roger-luo:
**No performance impact on native builds.** Every change in this PR is wasm-only — a `cfg(not(target_arch = "wasm32"))` dependency-table move or code gate, or the new `FingerprintHasher` alias that resolves to the existing `GxHasher` on native — so the native dependency + feature tree is byte-for-byte identical to `main` and the compiled native code is unchanged. A `ppvm-tableau-sum` `storage_compare` comparison against a `main` baseline confirmed this empirically: the deltas were pure measurement noise (the `vec` path, which this PR cannot touch, swung +8.9% then −3.2% across two runs of the *identical* binary).

Roger-luo:
I think WASM is fully supported now.

cc: @rafaelha

rafaelha:
Confirming that WASM compilation works for me now

PR #145, opened 2026-06-23
refactor(traits): device-portable PauliBackend abstraction (RFC scaffold)

Roger-luo:
## What & why

Scaffold + design RFC for a **de-biased map abstraction** so a CUDA-backed map can be a drop-in `Config::Map`, giving `CuPauliSum = PauliSum<CudaConfig>` — **one** Rust type, no parallel device type.

The current `ACMap` family is biased toward host (CPU) semantics in two ways that block a GPU backend:

1. **Host closures** — every transform takes `Fn(&W, &V) -> …` (`map_add_assign`, `map_insert*`, `scale`, `retain`, `contains_with`). The gate math lives in these closures in `ppvm-pauli-sum` (`sum/clifford.rs`, `rot1.rs`, `rot2.rs`). You can't ship an arbitrary Rust closure to a CUDA kernel.
2. **Host-reference iteration** — `ACMapIter::Item = (&W, &V)`; a device-resident map can't hand out `&V`, and per-term host iteration would erase the GPU win.

`PauliBackend` instead **reifies** the (small, closed) Pauli-propagation vocabulary — Clifford 1q/2q, Pauli rotations, `scale`/`merge`/`truncate`/`overlap` — as data (gate + qubit indices + scalar coeffs). Each backend supplies a CPU loop or a GPU kernel.

## What's in this PR (additive — nothing rewired yet)

- `docs/design/device-portable-pauli-backend.md` — problem, design, CPU-vs-CUDA impl table, 7-step migration plan, open questions.
- `ppvm-traits`: `PauliBackend` trait + `Clifford1q`/`Clifford2q` enums + `PackedWord` (canonical `(u64,u64)` encoding, shared with the CUDA backend).
- `reference::HashMapBackend` — packed-key CPU reference proving the shape: Clifford-1q (10 gates) + `scale_all`/`merge`/`truncate_abs`/`overlap`/`export` implemented and **tested (6 tests)**; rotations + 2q Cliffords marked `ITERATE`.

`cargo test -p ppvm-traits` and `cargo check --workspace` are green; `main` is untouched.

## How to iterate

This is intentionally a scaffold so it can be picked up async. The migration steps and open questions (Complex coeffs on cuco's 8-byte path, where to hoist `levi_civita`/`comm_2`, truncation reification) are itemized in the design doc — grab a row.

### DIFF
--- a/crates/ppvm-traits/src/traits/backend.rs
+++ b/crates/ppvm-traits/src/traits/backend.rs
@@ -1,0 +1,444 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Device-portable Pauli backend abstraction.
+//!
+//! This is the **de-biased** replacement for the closure-driven [`ACMap`] trait
+//! family. See `docs/design/device-portable-pauli-backend.md` for the full
+//! rationale. The short version:
+//!
+//! * [`ACMap`] expresses every transformation as a host Rust closure
+//!   (`Fn(&W, &V) -> …`) and iterates by host reference (`(&W, &V)`). Neither can
+//!   cross to a CUDA kernel / device-resident map.
+//! * [`PauliBackend`] instead exposes a **fixed, reified vocabulary** of Pauli
+//!   propagation operations (gate identity + qubit indices + scalar coeffs). Each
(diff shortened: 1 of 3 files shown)


Roger-luo:
hmm the problem is real here, but the abstraction is not the right shape, I prob need to redesign this myself instead of asking claude doing it... let me open an issue first

PR #131, opened 2026-06-17
feat(stim): X/Y-basis measure/reset and rec[-k] feed-forward gates

rafaelha:
## Summary

Extends the Stim subset accepted by `GeneralizedTableau.run()` (and the `StimProgram` parser/executor) with two features that were previously rejected:

1. **X/Y-basis measurement & reset** — `MX`, `MY`, `MRX`, `MRY`, `RX`, `RY`.
2. **Classically-controlled feed-forward** — controlled Paulis whose control is a measurement record, e.g. `CNOT rec[-1] 1`, plus `CX`/`CY`/`CZ rec[-k]`.

Both were singled out as gaps in the support report; `M !0`, `sweep[k]`, `MPP`/`MXX`/`MYY`/`MZZ`, and `CORRELATED_ERROR` remain out of scope (and now have corpus fixtures documenting the boundary).

## Approach (mirrors upstream Stim)

**X/Y-basis gates** are implemented via Stim's basis-change decomposition around the existing, well-tested Z-basis machinery, so readout-noise and RNG-draw shape stay identical to `M`/`MR`:

- `MX = H · M · H`, `MY = S_DAG · H · M · H · S`
- `RX = R · H` (→ `|+⟩`), `RY = R · H · S` (→ `|i⟩`)
- `MRX`/`MRY` add the reset to the measured eigenstate.

`MX`/`MY`/`MRX`/`MRY` already parsed (rejected at the executor); `RX`/`RY` are new `GateName` variants (`ResetX`/`ResetY`).

**Feed-forward** required a typed gate operand: a new `Target` enum (`Qubit(usize)` | `Rec(k)`) replaces the qubit-only `Vec<usize>` on the *gate* target path, threaded through parser → extended AST → executor. The control bit is looked up in the running measurement record and the target Pauli applied iff it is 1 — matching `TableauSimulator::single_cx`/`single_cy`. `prepare` rejects `rec[-k]` on non-controlled gates and in the Pauli-*target* slot ("measurement record editing is not supported").

## Tests
- New `crates/ppvm-stim/tests/xy_and_feedforward.rs` — quantum semantics (eigenstate prep, feed-forward on/off, lookback depth, REPEAT).
- New `crates/stim-parser/tests/record_targets.rs` — `rec[-k]` parsing, `RX`/`RY`, round-trip display, rejection of `rec[0]`/`rec[1]`/`rec[]`.
[...]

PR #147, opened 2026-06-23
feat(stim): X/Y-basis measure/reset, rec[-k] feed-forward, MPP, and native T/T_DAG

david-pl:
## Summary

Extends the Stim subset accepted by `GeneralizedTableau` so the magic-state cultivation circuit `crates/ppvm-stim/tests/data/cultivation_d5.stim` runs end-to-end. Two commits:

1. **Port of #131** (authored by @rafaelha) — X/Y-basis measure/reset (`MX`, `MY`, `MRX`, `MRY`, `RX`, `RY`) and classically-controlled feed-forward (`CX`/`CNOT`/`CY`/`CZ rec[-k]`), carried over onto current `main`.
2. **MPP + native `T`/`T_DAG`** (new) — multi-qubit Pauli-product measurement and the bare `T`/`T_DAG` mnemonics the cultivation circuit uses.

#131 targeted the unmerged `stim-mirror-api` branch, so this is a port (not a cherry-pick): `prepare`→`validate`, `execute_prepared`→`execute_validated`, accessor methods instead of direct `measurement_record` access, and the batched `_many` fast paths preserved for gates with no record control.

## X/Y-basis & feed-forward (commit 1, @rafaelha)

X/Y-basis gates use Stim's basis-change decomposition around the existing Z-basis machinery (`MX = H·M·H`, `MY = S_DAG·H·M·H·S`, `RX = R·H`, `RY = R·H·S`), so readout-noise / RNG-draw shape stays identical to `M`/`MR`. Feed-forward adds a typed gate operand: a new `Target` enum (`Qubit` | `Rec`) replaces the qubit-only `Vec<usize>` on the gate path, threaded parser → extended AST → executor. The control bit is read from the running measurement record and the Pauli applied iff it is 1, mirroring `TableauSimulator::single_cx`. `validate` rejects `rec[-k]` on non-controlled gates and in the Pauli-target slot.

## MPP & native T/T_DAG (commit 2)

- **MPP**: new `PauliFactor`/`PauliAxis` AST and a dedicated `Mpp` instruction variant carry the `X0*Y1*Z2`-style products (their own parse path, since targets are Pauli factors not qubit indices). The executor measures each product **non-destructively** via Stim's basis-change + CX-ladder gadget: rotate each factor's axis to Z (`X = H·Z·H`, `Y = (S H)·Z·(H S_DAG)`), a CX ladder onto the first qubit folds `Z_0 … Z_{m-1}` into a single `Z_0`, measure that qubit, then undo the ladder and basis changes so only the product operator is projected. One result per product; optional readout-noise arg.
- **Native `T`/`T_DAG`**: cultivation uses the bare mnemonics rather than the `S[T]`/`S_DAG[T]` sugar. Added as `GateName` variants that `interpret` lowers to the existing `ExtendedInstruction::T`/`TDag`, sharing the tableau's T-gate path.

## cultivation_d5.stim

Parses, validates, and executes end-to-end: **112 measurements** (36 `M` + 57 `MX` + 19 `MPP` products). Runtime ~11ms release / ~240ms debug — only 11 `T`/`T_DAG` gates. Covered by `tests/cultivation.rs` and wired into the corpus.

## Tests

- `stim-parser/tests/{record_targets,mpp}.rs` — `rec[-k]`/`RX`/`RY` parsing, MPP product parsing + round-trip + rejection.
- `ppvm-stim/tests/{xy_and_feedforward,mpp,cultivation}.rs` — quantum semantics (eigenstate prep, feed-forward on/off, lookback depth, REPEAT; MPP parity / Bell eigenvalues / mixed-basis / non-destructiveness) and the full cultivation run.
- Updated `validate.rs`, `stim_corpus.rs`, corpus fixtures, `proptest_ast.rs`.
- Python binding tests for X/Y, feed-forward, and MPP in `test_stim.py`.

## Checks

- `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `cargo check --workspace --all-targets` — clean.
- Full Rust `stim-parser` + `ppvm-stim` suites pass; full Python suite (191 tests) passes after a native rebuild.

Still out of scope (and documented by corpus fixtures): `M !0` inverted targets, `MXX`/`MYY`/`MZZ`.


david-pl:
FYI, @rafaelha @Roger-luo: turns out we do need this if we want to reproduce the benchmarks from the clifft paper shown in table 1. For that, we also need MPP. And I also added a T and T_DAG instruction here, which acts the same way as the tagged S[T]. I guess it doesn't hurt to have those synonyms.

rafaelha:
Amazing, this looks great! And I will be using this for the playground as well.

One small subtlety for MPP. When paulis are repetaed, i.e. `MPP X0*X0`, stim simplifies this as `MPP I` (and always returns 0). Or `MPP X0*X0*Z0` becomes `MPP Z0`.

For something like `MPP Z0*X0`, stim will raise `ValueError: Acted on an anti-Hermitian operator...` because Z0*X0=iY0 is non-hermitian.

ppvm does not currently treat this edge case correctly. The simplest way would be to disallow repeated Paulis on the same index. Or you could do what stim does.

image

crates/ppvm-stim/src/executor.rs
```diff
@@ -11,11 +11,105 @@ use std::fmt::Debug;
...
+/// takes the per-pair feed-forward path rather than the batched fast path.
+fn has_record_control(targets: &[Target]) -> bool {
+    targets.iter().any(|t| matches!(t, Target::Rec(_)))
+}
+
+/// Resolve a measurement-record lookback `rec[-k]` against the running record.
+/// `k == 1` is the most recent measurement. A lost-qubit measurement (`None`)
+/// or an out-of-range lookback resolves to `false` (no feed-forward applied).
+#[inline]
+fn record_bit(record: &[Option<bool>], k: usize) -> bool {
```
rafaelha, line 62: [P1] Out-of-range `rec[-k]` controls silently no-op instead of erroring. [executor.rs (line 60)](/Users/rafaelhaenel/Documents/quera/ppvm/crates/ppvm-stim/src/executor.rs:60) uses `unwrap_or(false)` when the lookback is before the measurement record. Stim raises `IndexError` for `CX rec[-1] 0` with no prior measurement and for `M 0; CX rec[-2] 0`; ppvm returns `Ok` and applies no correction. That can hide invalid circuits as valid no-ops.

crates/stim-parser/src/extended/interpret.rs
```diff
@@ -90,12 +101,21 @@ fn interpret_gate(
...
     args: Vec<f64>,
-    targets: Vec<usize>,
+    targets: Vec<Target>,
     line: usize,
 ) -> Result<ExtendedInstruction, ExtendedParseError> {
     use GateName::*;
 
     match name {
+        // Native T / T_DAG mnemonics lower to the same sugar as `S[T]` / `S_DAG[T]`.
+        T => Ok(ExtendedInstruction::T {
```
rafaelha, line 111: [P2] `parse_extended` can panic before validation for `rec[-k]` on sugar gates. [interpret.rs (line 111)](/Users/rafaelhaenel/Documents/quera/ppvm/crates/stim-parser/src/extended/interpret.rs:111) lowers `T`, `T_DAG`, `S[T]`, and tagged `I[...]` via `qubit_targets`, which expects every target to be a qubit at [interpret.rs (line 343)](/Users/rafaelhaenel/Documents/quera/ppvm/crates/stim-parser/src/extended/interpret.rs:343). Since the parser now accepts `rec[-k]` on all gates, `parse_extended("T rec[-1]")` and `parse_extended("S[T] rec[-1]")` panic instead of returning a parse/validation error.

rafaelha (commented):
Review comments on uncovered edge cases.

Roger-luo:
I think I want to get this PR in, but I'd like to refactor the Steam code executor a bit first. Previously, we were just using a simple loop to run a Steam text file, but it is getting more complicated now. Since it doesn't have any abstractions, the code is becoming a little bit cumbersome to me at the moment.

PR #157, opened 2026-06-24
perf(tableau-sum): 5.04x faster GeneralizedTableauSum build (lazy branch materialization + faster fingerprinting)

david-pl:
## Summary

Autotune session optimizing `GeneralizedTableauSum` **build time** on the `msd-noisy` workload (magic-state-distillation circuit with single-qubit depolarizing + loss noise).

**Build time: 2620 ms → 520 ms — a 5.04× speedup.** Sampling throughput is unchanged (~22 µs/shot) and the result is **bit-identical**: branch count stays `2025`, `sum_p2 = 0.725135705447`, and the top-5 probabilities match to the recorded reference. All changes live in `crates/ppvm-tableau-sum/src/`.

## What landed

| # | Change | Build | Step | Cumulative |
|---|--------|-------|------|------------|
| 0 | baseline | 2620 ms | — | 1.0× |
| 1 | **Lazy branch materialization** — noise no longer deep-clones a ~7 KB tableau per branch; it derives each branch's fingerprint cheaply and only clones survivors (~85% of branches merge/drop) | 958 ms | 2.73× | 2.73× |
| 2 | **Bulk word-hash** — gather a row's words once + a single `gxhash64`, instead of ~340 tiny hash writes | 573 ms | 1.67× | 4.57× |
| 3 | **Precompute per-row masks** — build the splitmix mask table once per op, not per-row-per-entry | 552 ms | 1.04× | 4.75× |
| 4 | **Direct-word column reads** — replace `bitvec` per-bit indexing with direct storage-word access in the noise hot loops | 520 ms | 1.04× | 5.04× |

The dominant lever was #1: profiling showed `fork`/clone was 47% of build (32% raw `memmove`); after lazy materialization it dropped to ~0.1%.

### Detail on #1 (lazy materialization)
`GeneralizedTableau` X/Y/Z gates only flip per-row sign bits and leave words/coefficients/`is_lost` identical to the parent; loss only sets one `is_lost` bit. So a branch's fingerprint is derivable from the parent without cloning. This PR adds `BranchMutation` + `structurally_equal_mutated` + `apply_branch_mutation` in `storage/mod.rs`, and an `insert_or_merge_mutated_branches` trait method (`VecStorage` does the lazy/index-based path; `MapStorage` keeps an eager-clone fallback). `loss_channel` and `pauli_error` emit virtual branches and only clone the survivors. The two-qubit / correlated-loss / reset-loss paths are left eager (not exercised by `msd-noisy`).

## One discarded experiment (kept for the record)
A scalar single-pass word hash (replacing the gather + `gxhash`) **regressed** to 921 ms and was reverted. This confirms the fingerprint rebuild is SIMD-throughput-bound — full re-hashing via `gxhash` is near-optimal, so the only remaining lever is to avoid re-hashing entirely (incremental/Zobrist fingerprinting). That is **not** in this PR.

## Where the time goes now (520 ms plateau)
- fingerprint rebuild ~37% (word-hash gather 22% + `phase_loss_hash` 14.5%) — proven SIMD-compute-bound
- noise branch-building ~21%
- inherent per-entry gate application ~23% (cz 10.5%, sqrt_* 12.7%)

The next lever would be cell-level incremental ("Zobrist") fingerprinting maintained through the Clifford gates to skip the rebuild (~10–20% est.). It is deliberately **not** included — it's a large, clever change (new hash scheme + per-gate delta logic incl. cz cross-terms) that runs against the repo's "readability over cleverness" preference, for a modest further gain.

## Testing
- `cargo test -p ppvm-tableau-sum` — all crate tests pass
- Accuracy is guarded by `test_word_fingerprint_cache_stays_consistent` plus the `branches == 2025` invariant in the bench harness, so any fingerprint drift fails loudly.

## Notes for reviewers
- `crates/ppvm-tableau-sum/examples/msd-noisy-bench.rs` is the deterministic, seeded measurement harness used throughout (median-of-5 build time + accuracy fingerprint). It's included so the numbers are reproducible; drop it if you'd rather not carry a bench example.
- The session's working notes (per-iteration log, metric ledger, prompt records) were kept out of the PR — they're summarized above.


Roger-luo:
@david-pl I want to merge this pr so I can start refactor the trait system

david-pl:
@Roger-luo I went through it again and I think it's in an okay shape. So, fine to merge since you want to run cleanup anyway.

Roger-luo:
Let's merge this. I want to do a small refactor on the tree system afterwards anyways. It's going to be easier to think about new abstractions after merging.

PR #159, opened 2026-06-24
feat(stim-parser): native R_X/R_Y/R_Z/U3 rotation gates + require *pi in tags

david-pl:
## Summary

Lets `stim-parser` / `ppvm-stim` natively ingest the clifft/tsim rotation
vocabulary, so circuits using `R_X`/`R_Y`/`R_Z`/`U3` no longer need tsim's
`shorthand_to_stim` rewrite before ppvm can run them.

Two commits:

### 1. `feat(stim-parser): parse bare R_X/R_Y/R_Z/U3 rotation gates`
- Register `R_X`/`R_Y`/`R_Z` (`Exact(1)`) and `U3` (`Exact(3)`) in the
  instruction table; lower them to the existing `Rotation`/`U3` extended
  variants. No executor/validator changes needed downstream.
- The bare argument is in **half-turns** (clifft convention:
  `R_Z(a) = exp(-i*a*pi/2*Z)`, `gate_data.h`/`frontend.cc`), so lowering
  multiplies by π — making `R_Z(a)` identical to `I[R_Z(theta=a*pi)]`.
  Verified empirically: clifft and ppvm rotation statistics match to
  sampling noise across several angles.
- ppvm-stim's exhaustive `GateName` matches gain the new variants; they
  always lower away before execution, so they join `T`/`T_DAG` as
  `Ok`-in-validate / `unreachable!`-in-executor.

### 2. `feat(stim-parser)!: require *pi in rotation/U3 tag angles`
- Mirror tsim's parametric-tag convention: tag angles must be written as
  `<n>*pi` (`I[R_Z(theta=0.5*pi)]`); a bare number (`theta=0.5`) is now
  rejected with `invalid-tag`. tsim's tag parser requires the `*pi` literal
  and reads the coefficient as half-turns; ppvm refuses the ambiguous bare
  form rather than silently treating it as radians.
- `TagParam::Named` gains a `had_pi` flag (new `pi_expr_flagged` grammar
  rule); `exact_named_params` enforces it; the printer re-emits `*pi`
  (`theta={theta/pi}*pi`) so print→parse stays a fixpoint.
- The bare *gate* form `R_Z(0.5)` is unchanged — only the bracketed tag is
  tightened, matching tsim (shorthand `R_Z(a)` vs canonical `theta=a*pi`).

## Background

clifft uses half-turn units; tsim's canonical tag is `theta=a*pi` (it
stores the coefficient as an exact `Fraction`). ppvm's tag stored radians
and was the outlier — it coincided with the ecosystem only on the `*pi`
form. These commits make the bare mnemonic clifft-faithful and tighten the
tag to the `*pi` convention.

## Test plan
- `cargo test -p stim-parser -p ppvm-stim` — all 25 test binaries pass.
- TDD throughout (each test watched failing first); clippy/fmt/license hooks pass.
- `cargo build --workspace` clean.

## Notes
- One `ppvm-python` fixpoint test was updated to the `*pi` form. It only
  runs against the compiled wheel, which still ships old `ppvm 0.1.0` — the
  feature reaches Python only after `ppvm-python-native` is rebuilt with
  maturin (not done here).
- `SPP[T]`/`SPP_DAG[T]` (Pauli-product rotation, from `TPP`) remain
  unsupported and out of scope.

### DIFF
--- a/crates/ppvm-stim/src/executor.rs
+++ b/crates/ppvm-stim/src/executor.rs
@@ -644,6 +644,11 @@ pub fn execute_validated<T, I, C>(
                 GateName::T | GateName::TDag => {
                     unreachable!("T/T_DAG are lowered to ExtendedInstruction::T/TDag by interpret")
                 }
+                GateName::RotX | GateName::RotY | GateName::RotZ | GateName::U3 => {
+                    unreachable!(
+                        "R_X/R_Y/R_Z/U3 are lowered to ExtendedInstruction::Rotation/U3 by the parser"
+                    )
+                }
             },
             ExtendedInstruction::T { targets, .. } => targets.iter().for_each(|&q| tab.t(q)),
             ExtendedInstruction::TDag { targets, .. } => {
--- a/crates/ppvm-stim/src/validate.rs
+++ b/crates/ppvm-stim/src/validate.rs
@@ -123,8 +123,8 @@ fn check_gate_supported(name: GateName, line: usize) -> Result<(), ExecError> {
     use GateName::*;
     match name {
         Reset | ResetZ | ResetX | ResetY | X | Y | Z | H | HXZ | S | SqrtZ | SDag | SqrtZDag
-        | SqrtX | SqrtXDag | SqrtY | SqrtYDag | T | TDag | Identity | CX | ZCX | CNot | CY
-        | ZCY | CZ | ZCZ => Ok(()),
+        | SqrtX | SqrtXDag | SqrtY | SqrtYDag | T | TDag | Identity | RotX | RotY | RotZ | U3
+        | CX | ZCX | CNot | CY | ZCY | CZ | ZCZ => Ok(()),
         Swap | ISwap | ISwapDag | SqrtXX | SqrtYY | SqrtZZ | CXSwap | SwapCX | XCX | XCY | XCZ
         | YCX | YCY | YCZ | CXYZ | CZYX | HXY | HYZ => Err(ExecError::Unsupported {
             name: name.canonical_name().to_string(),
--- a/crates/ppvm-stim/tests/executor.rs
+++ b/crates/ppvm-stim/tests/executor.rs
@@ -46,7 +46,7 @@ fn rx_pi_flips_qubit() {
 
 #[test]
 fn u3_pi_flip_via_y_axis() {
-    let (results, _) = run("I[U3(theta=1.0*pi, phi=0.0, lambda=0.0)] 0\nM 0", 1);
+    let (results, _) = run("I[U3(theta=1.0*pi, phi=0.0*pi, lambda=0.0*pi)] 0\nM 0", 1);
     assert_eq!(results, vec![Some(true)]);
 }
 
--- a/crates/ppvm-stim/tests/executor.rs
+++ b/crates/ppvm-stim/tests/executor.rs
@@ -51,6 +51,24 @@ fn u3_pi_flip_via_y_axis() {
 }
 
 #[test]
+fn u3_all_angles_nonzero_exercises_phi_lambda() {
+    // U3(theta=pi, phi=pi/2, lambda=pi/2) == Y (clifft Rz(phi)Ry(theta)Rz(lambda)),
+    // so H·U3·H == H·Y·H == -Y and |0> -> |1> deterministically. The H frame makes
+    // the outcome sensitive to phi *and* lambda (drop or mis-scale either and P(1)
+    // collapses to ~0.5). Half-turn args 1.0/0.5/0.5 each get *pi at lowering.
+    let tag = run(
+        "H 0\nI[U3(theta=1.0*pi, phi=0.5*pi, lambda=0.5*pi)] 0\nH 0\nM 0",
+        1,
+    );
+    assert_eq!(tag.0, vec![Some(true)]);
+
+    // Bare half-turn form must agree; this also pins the `* pi` scaling (an
+    // unscaled bare U3 would read 1.0/0.5/0.5 as radians and P(1) drops to ~0.4).
+    let bare = run("H 0\nU3(1.0, 0.5, 0.5) 0\nH 0\nM 0", 1);
+    assert_eq!(bare.0, vec![Some(true)]);
+}
+
+#[test]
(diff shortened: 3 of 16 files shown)


david-pl:
FYI @rafaelha @Roger-luo: adds some additional non-Clifford instructions that tsim and clifft support. With this, we support everything but TPP and SPP, I think.

Roger-luo:
## Review: edge cases + a pushed fix

Reviewed the PR for unhandled edge cases. Found one confirmed correctness bug and pushed a fix to this branch (`a97599c3`); two other items are flagged below for your call.

### 🔴 Fixed in `a97599c3` — bare rotation/U3 args silently double-applied π

Bare rotation/U3 args are in half-turns and get `× π` at lowering (`R_Z(0.5)` == `I[R_Z(theta=0.5*pi)]`). But args flow through the generic `pi_expr` parser, so the `*pi`/`pi` form was accepted **and then multiplied by π again**:

```
R_Z(0.5*pi) 0   ->  theta = 4.9348…  (= 0.5·π², not 0.5·π)
R_Z(pi) 0       ->  theta = π²
```

No diagnostic — the inverse of the ambiguity commit #2's `*pi` tag requirement was added to prevent. A user copying the `*pi` coefficient from the canonical tag form into the bare gate gets a silently wrong angle.

**Fix (targeted, not global):** `args_block` now records a `had_pi` flag (reusing the existing `pi_expr_flagged`), and the validator rejects a `*pi` arg **only** on the bare `R_X`/`R_Y`/`R_Z`/`U3` mnemonics with a clear `half-turn-arg` error. Crucially, `pi`-expressions stay valid in args for every other instruction — `noise.rs` documents that as an intentional `stim.rs`-compat feature (`X_ERROR(0.5*pi)`), so a global removal would have been wrong.

```
R_Z(0.5) 0         -> OK
R_Z(0.5*pi) 0      -> ERROR 'R_Z' takes angle arguments in half-turns; remove the '*pi' suffix …
U3(0.5*pi,0,0) 0   -> ERROR (same)
X_ERROR(0.5*pi) 0  -> OK  (compat preserved)
```

Added 5 regression tests (grammar / validate-unit / lower-e2e). `cargo test -p stim-parser -p ppvm-stim` green, `cargo build --workspace` clean, all pre-commit hooks pass.

### 🟡 For your call — U3 `phi`/`lambda` half-turn convention

Commit #1 multiplies all three U3 angles by π, but the empirical verification cited in the PR covers the single-axis **rotations**; the only U3 execution test uses `theta=1.0*pi` with `phi=lambda=0`, so the half-turn scaling of `phi`/`lambda` is never exercised with nonzero values. Worth either confirming clifft's U3 uses half-turns for all three angles, or adding a U3 test with nonzero `phi`/`lambda`. (Didn't touch this — I can't verify against clifft, and a test here would just encode a guess.)

### 🟢 Confirm intent — bare `T`/`T_DAG` now reject tags

Commit #1 adds tag rejection to the bare `T`/`T_DAG` arm (previously tags were silently dropped). Correct and consistent with the new rotation behavior, just slightly broader than the "rotation vocabulary" scope in the summary — flagging so it's a deliberate choice.

### ✅ Checked and fine (no action)

- **Printer fixpoint:** the new divide-on-print / multiply-on-parse `*pi` cycle is stable — simulated the full pipeline over 300k random angles, settles on the first print; the AST proptest's exact-eq property holds for all `float_lit()` values.
- **Arity soundness:** the `unreachable!("arg count is fixed…")` in lowering is sound — `Exact(1)`/`Exact(3)` is enforced in validate before lowering.
- Record targets / positional tag params / empty tag blocks on the new gates are all handled.

Roger-luo:
## Addressed Copilot's printer note (`print/mod.rs`) — `b7e282fe`

The rotation/U3 printer emitted `theta/PI` directly, so the division's rounding tail leaked into the output — e.g. `theta=0.76*pi` printed as `theta=0.7599999999999999*pi`. ~10% of ordinary decimal angles were affected.

**Fix:** a new `pi_coeff(value)` returns the *shortest decimal `c` whose `c * PI` recovers the stored radians bit-for-bit*, falling back to `value / PI` when no exact short form exists. Because acceptance requires **exact** equality, the output is both clean and round-trip-lossless:

```
I[R_Z(theta=0.76*pi)] 0   ->  I[R_Z(theta=0.76*pi)] 0     (was 0.7599999999999999*pi)
R_Z(0.19) 0               ->  I[R_Z(theta=0.19*pi)] 0
I[U3(theta=0.34*pi, …)]   ->  unchanged, no tail
```

**Round-trip safety** (the key constraint): since `c` is only accepted when `c*PI == theta` exactly, `parse → print` stays lossless and `print → parse → print` remains a byte-for-byte fixpoint. Covered by:
- a new property test (`rotation_pi_coeff_round_trips`) over arbitrary decimal coefficients asserting **exact theta recovery + fixpoint + no rounding tail** (stress-run at 20k cases),
- a unit test covering tags, bare gates, and U3,
- and the existing fixpoint / AST-roundtrip proptests, all still green.

`cargo test -p stim-parser -p ppvm-stim`, `cargo clippy -- -D warnings`, and fmt all pass.

david-pl:
Addressed the 🟡 U3 `phi`/`lambda` item (`8c6e9bd7`).

Confirmed the convention against clifft (`src/clifft/frontend/frontend.cc`): `R_Z(alpha) = exp(-i·alpha·π/2·Z)` (half-turns), and `U3(theta, phi, lambda) = R_Z(phi)·R_Y(theta)·R_Z(lambda)` feeds all three angles through the same `trace_rz` (with a `(phi+lambda)·π/2` global phase) — so all three U3 angles are half-turns, and the `× π`-on-all-three lowering is correct.

Added `u3_all_angles_nonzero_exercises_phi_lambda` to `crates/ppvm-stim/tests/executor.rs`. It uses `U3(θ=π, φ=π/2, λ=π/2) == Y`, framed as `H · U3 · H` so `|0> → |1>` deterministically. The H frame makes the outcome phase-sensitive: dropping or mis-scaling `phi` or `lambda` collapses `P(1)` from `1.0` to `~0.4–0.5` (verified numerically against the clifft matrix). Covers both the tag form and the bare half-turn form, the latter also pinning the `* pi` scale. `cargo test -p stim-parser -p ppvm-stim` green.

PR #163, opened 2026-06-26
feat(stim-parser)!: require *pi in rotation/U3 tag angles (no bare rotation gates)

david-pl:
## Supersedes #159

This PR replaces #159 (`david/parse-bare-rotations`). It keeps **only** the tag-parsing tightening from that PR and **drops the bare rotation-gate support**, matching **tsim** semantics rather than **clifft**'s bare half-turn shorthand.

#159 should be closed in favor of this one.

## What changes

- **No bare `R_X` / `R_Y` / `R_Z` / `U3` gates.** The bare clifft-style mnemonics (`R_Z(0.5) 0`, `U3(a,b,c) 0`) are *not* registered, so they are rejected as `unknown instruction` — the same as on `main`. Only tsim's canonical tagged form is accepted.
- **`*pi` is strictly required in rotation/U3 tag angles.** Mirroring tsim's parametric-tag convention, angles must be written in half-turns as `<n>*pi` (e.g. `I[R_Z(theta=0.5*pi)]`). A bare number (`I[R_Z(theta=0.5)]`) is now rejected with an `invalid-tag` diagnostic rather than silently read as radians.
- **Printer re-emits the clean `<c>*pi` form** with the shortest exact coefficient (no `0.7599999999999999*pi` rounding tail), keeping `parse → print` a byte-for-byte fixpoint.
- **`T` / `T_DAG` bare gates now reject tags** (`T[foo] 0` previously parsed and silently dropped the tag) — an orthogonal tightening included here.

### Behavior

| Input | Result |
|---|---|
| `I[R_Z(theta=0.5*pi)] 0` | ✅ accepted, round-trips |
| `I[R_Z(theta=0.5)] 0` | ❌ `parameter 'theta' must be written as <n>*pi (half-turns)` |
| `R_Z(0.5) 0` | ❌ `unknown instruction 'R_Z'` |
| `U3(0.5, 1.0, 1.5) 0` | ❌ `unknown instruction 'U3'` |
| `I[U3(theta=0.3*pi, …)] 0` | ✅ accepted, round-trips |

## Commits

1. `feat(stim-parser)!: require *pi in rotation/U3 tag angles`
2. `fix(stim-parser): reject tags on bare T/T_DAG gates`
3. `fix(stim-parser): print shortest exact *pi coefficient (no rounding tail)`
4. `test(ppvm-stim): U3 with all angles nonzero exercises phi/lambda half-turns`

## Testing

- `cargo test -p stim-parser -p ppvm-stim` — all green.
- `pytest test/test_stim_api.py` — 22 passed (native module rebuilt).


david-pl:
FYI, @Roger-luo I checked again and these are the semantics as emitted by bloqade and parsed by tsim.

Roger-luo:
can we also support the syntax `0.5pi` while also supporting `0.5*pi`? just for better readability without creating any ambiguity

Roger-luo (approved):

PR #164, opened 2026-06-30
feat(tableau): word-fused cz_block and faster Python MSD path

david-pl:
## Summary

Speeds up the Python `GeneralizedTableau` MSD path with a new word-fused CZ primitive and lower-overhead gate dispatch, and adds benchmarks for both languages.

The 85-qubit MSD circuit **build** drops from ~124µs to ~62µs. The **full shot** (build + readout) is ~92µs versus the Rust fused bench's ~56µs — still **~1.6×**, not parity. The remaining gap is per-call Python dispatch, the scattered encode CZs (which use `cz_many` in Rust too), and the measurement loop (real work both languages pay).

### New `cz_block` fused CZ (Rust + Python)
- `GeneralizedTableau::cz_block(control_base, target_base, count)` in `ppvm-tableau`: applies CZ to constant-offset pairs `(control_base+i, target_base+i)` and **splits the run at `u64`-word boundaries internally**, dispatching to the existing `cz_block_pairs` / `cz_block_pairs_cross_word` kernels. Callers use plain qubit indices — no word/bit arithmetic.
- Exposed through the PyO3 interface and the Python `GeneralizedTableau` wrapper (`+ _core.pyi` stub).
- This is the lever for the build speedup: the contiguous cross-block CZ layers go from per-pair `cz_many` (O(n) per pair) to word-parallel block ops (O(n/64) per block). The scattered encode CZs stay on `cz_many` (no contiguous structure to fuse — same as the Rust bench).

### Faster Python gate dispatch
- `mixins`: pass gate targets straight to the native layer — PyO3 extracts `Vec<usize>` directly from lists/tuples/ranges/ndarrays — instead of rebuilding a list with a per-element `int()` on every call. Concrete-type fast paths in `_is_sequence` avoid the slow ABC `isinstance(obj, Iterable)`.
- Measurement: convert raw outcome codes through a shared `_BY_VALUE` tuple lookup rather than the per-element `MeasurementResult()` IntEnum constructor (deduplicated with `GeneralizedTableauSum`).

### Benchmarks
- New pytest-benchmark `test/benchmarks/test_msd.py`: the 85-qubit MSD circuit with splatted gates, build-only and build+measure arms (smoke-tests by default; `--benchmark-enable` to time).
- `tableau-msd-fused.rs` now calls `cz_block` with qubit indices instead of the hand-split word/bit calls — runtime unchanged (verified with a criterion A/B baseline: within noise).

## Verification
- `ppvm-tableau` lib tests pass, including a new `cz_block` cross-word + reversed-bases equivalence test.
- 204 Python tests pass (covers list/tuple/range/ndarray/`np.int64` target forms).
- `cz_block` proven equivalent to per-pair `cz` on coefficients.


david-pl:
FYI, @Roger-luo this removes some regressions introduced when matching STIM parity (`_normalize_targets` was slow).

Roger-luo (approved):

PR #174, opened 2026-07-08
feat(ppvm-vihaco): add the composite machine and VM

david-pl:
## Merge order

Part of the #168 split. This PR targets **#173** (`david/42.2-circuit-component`) and sits at the top of the vihaco line:

1. RotXY / R gate (#170) → `main` ✅ merged
2. tableau expectation/reset (#172) → `main` ✅ merged
3. vihaco-circuit-isa (#169) → `main` (open; carries #170 + #172 via a `main` merge)
4. native `.pyi` stubs (#171) → `main` (open; independent)
5. circuit component (#173) → #169
6. **composite machine + VM (this PR)** → #173
7. CLI + TUI (#166) → this PR

## Summary

Completes the `ppvm-vihaco` crate on top of the circuit component (#173) — the composite machine that drives it plus the surrounding VM.

| File | What |
|------|------|
| `composite.rs` | the `PPVM` composite: vihaco machine (CPU + program loader) driving the circuit component, measurement/trace observers, backend selection from the device header, reset, single-instruction stepping |
| `syntax.rs` | `.sst` header parse / lower (`PPVMHeader`, `PPVMResolver`) |
| `bytecode.rs` | `.ssb` bytecode emit / load (magic-tagged) |
| `observable.rs` | observable-header parsing for the PauliSum backends |
| `shots.rs` | multi-shot execution (serial + optional `rayon`) |
| `lib.rs` | full module wiring + `load`/`run`/`dump`/`compile`/`parse` helpers + `PPVMModule` |
| `Cargo.toml` | adds `vihaco-cpu`, `log`, optional `rayon` feature |
| `tests/*.sst` + `sst_fixtures.rs` | 28 integration fixtures |

## `device_info` relocation resolves here

`composite.rs` no longer defines `PPVM_MAGIC` / `BackendKind` / `PPVMDeviceInfo` — it imports them from `device_info` (introduced in #173) and `pub use`-re-exports them, so the existing `crate::composite::{…}` paths in `lib.rs` / `bytecode.rs` / `syntax.rs` keep resolving with no other edits.

The machete-ignore added in #173 is dropped here — the full crate uses `chumsky` / `vihaco-parser-core` directly.

## Testing

`cargo test -p ppvm-vihaco` — **111 unit + 28 integration** tests, green with and without `--features rayon`. Full workspace builds; clippy + machete clean.

PR #166, opened 2026-07-01
Add a composable ratatui TUI to ppvm-cli

david-pl:
## Summary

Adds a composable [ratatui](https://ratatui.rs) TUI to `ppvm-cli`. Bare `ppvm` launches a unified single-screen REPL + step-debugger; `ppvm <file>` opens a program paused at the first instruction. The existing `run`/`dump`/`parse`/`debug` subcommands are unchanged.

- **New library crate `crates/ppvm-tui`** holds the terminal-agnostic pieces: an `AppState` owning a `ppvm_vihaco::composite::PPVM`, the command grammar, and four ratatui `Widget` view components (Program/State/Measurement-record/Command line).
- **`ppvm-cli` (bin)** owns only terminal setup (RAII restore on panic) and the blocking event loop.
- **Composability:** view components own no terminal and run no loop (`impl Widget for &…View`-style; `handle_key(KeyEvent) -> bool`), and deps are pinned to `ratatui`/`crossterm` `0.29.0` to match `stellarscope` so panels can dock together later.

### Features
- **REPL:** `device N` then gate ops (`h 0`, `cnot 0 1`, `rx 0 0.5`, `measure 0`, …); outcomes echo as `=> bits`.
- **Debugger:** `:load`, Enter = step, `:c` = continue to next breakpoint/end, `:reset`; authored `breakpoint`s pause.
- **Breakpoint gate injection:** at a breakpoint you can apply ad-hoc gates and resume — backed by making `apply_circuit_instruction` preserve the program counter (`ppvm-vihaco`).
- **Line editing + history:** movable cursor (←/→, Home/End, insert/Backspace/Delete) drawn via the terminal cursor; ↑/↓ command history with a draft stash.
- **`:help`** toggles a command-reference overlay (works in both modes).

## Base branch

Targets **`david/42-circuit-component`** (not `main`) — this branch was cut from it and depends on its `vihaco-circuit-isa` work. The merge commit here also pulls in that branch's `Add back eyre` fix (a pre-existing derive-macro build breakage), so the branch compiles.

## Design docs
- Spec: `docs/superpowers/specs/2026-07-01-cli-tui-design.md`
- Plan: `docs/superpowers/plans/2026-07-01-cli-tui.md`

## Testing
- `cargo test --workspace` — green (0 failures).
- `ppvm-tui`: 36 tests (command grammar, REPL/debug state transitions, breakpoint injection resume, cursor/history editing, help toggle, `TestBackend` render checks).
- `clippy --all-targets` clean.

[...]

PR #175, opened 2026-07-08
feat(ppvm-cli): add the ppvm command-line binary

david-pl:
## Merge order

Part of the #168 split. This PR targets **#174** (`david/42.3-composite`) and adds the CLI binary the TUI PR builds on:

1. RotXY / R gate (#170) → `main` ✅ merged
2. tableau expectation/reset (#172) → `main` ✅ merged
3. vihaco-circuit-isa (#169) → `main` (open)
4. native `.pyi` stubs (#171) → `main` (open; independent)
5. circuit component (#173) → #169
6. composite machine + VM (#174) → #173
7. **ppvm-cli binary (this PR)** → #174
8. ratatui TUI (#166) → this PR

## Summary

Extracts the `ppvm-cli` crate — the `ppvm` binary — from #168 onto the composite. This is the **base CLI without the TUI**: clap-based `run` / `dump` / `parse` / `debug` subcommands over `ppvm-vihaco`, plus example `.sst` programs. The ratatui TUI that layers on top of it is the follow-up PR (#166), which modifies this crate.

`ppvm-cli` was part of #168's implementation but had no extraction step of its own — the original split plan conflated it with the TUI PR. It slots in here between the composite and the TUI.

- Depends on `ppvm-vihaco` (with `rayon`), `clap`, `eyre`.
- **Excluded from the wasm32 CI build** (terminal binary, forces `rayon`), mirroring #168's exclusion — `ppvm-vihaco` itself still builds for wasm, so only `ppvm-cli` is excluded; the reusable engine stays covered.

## Testing

`cargo test -p ppvm-cli` — 23 pass. `ppvm run examples/ghz.sst` runs. clippy + machete clean; full workspace builds; `ppvm-vihaco` still builds for `wasm32-unknown-unknown`.

### DIFF
--- a/crates/ppvm-cli/src/commands.rs
+++ b/crates/ppvm-cli/src/commands.rs
@@ -1,0 +1,580 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use eyre::{Result, WrapErr};
+use ppvm_vihaco::composite::{PPVM, StepOutcome};
+use ppvm_vihaco::measurements::MeasurementResult;
+use ppvm_vihaco::shots::ShotRecord;
+use std::io::{BufRead, Write};
+use std::path::Path;
+
+/// Output format for `parse`.
+#[derive(Clone, Debug, clap::ValueEnum)]
+pub enum Format {
+    Pretty,
+    Debug,
+    Json,
+}
+
+/// Output format for the measurement record from `run`.
+#[derive(Clone, Debug, clap::ValueEnum)]
+pub enum MeasurementFormat {
+    /// One flat bit string per shot: `0`/`1`, lost qubit = `2`.
+    Bits,
+    /// Raw debug representation of all shots.
+    Debug,
+}
+
+pub fn run(
+    file: &str,
+    shots: usize,
+    seed: Option<u64>,
+    output: Option<&str>,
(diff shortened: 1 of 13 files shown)


crates/ppvm-cli/src/commands.rs
```diff
@@ -0,0 +1,580 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
```
Roger-luo, line 1: I think a lot of the impl in this file should be provided by vihaco too

cc: @robpatterson13

crates/ppvm-cli/src/commands.rs
```diff
@@ -0,0 +1,580 @@
...
+/// program can be stepped from the beginning.
+pub fn debug(file: &str, break_at_start: bool) -> Result<()> {
+    let stdin = std::io::stdin();
+    let mut input = stdin.lock();
+    let mut output = std::io::stdout();
+    debug_loop(file, break_at_start, &mut input, &mut output)
+}
+
+/// Core debugger loop, generic over its IO so it can be driven by tests.
+fn debug_loop(
```
Roger-luo, line 183: I think vihaco needs to have some new traits to property define what a "debug-able" component is and then provide default debugger impl from there

cc: @robpatterson13

Roger-luo (approved):

PR #185, opened 2026-07-17
TUI follow-up

david-pl:
* Adds a CPU stack view
* Makes state view configurable (and skips details for many qubits)
* Adds a tree view that displays the number of branches and branching history
* Makes REPL commands case insensitive


Roger-luo:
Given now we have a GUI, maybe we can aim for something simple for this PR.

PR #187, opened 2026-07-17
feat(tableau): peek observable expectation — string observables + loss handling

rafaelha:
## What changes

Adds a non-destructive observable-expectation *peek* to `GeneralizedTableau`, layered on top of the expectation machinery from #172/#176:

- **`crates/ppvm-tableau/src/observable.rs`** (new): parsing of Pauli-observable strings. Two notations, each with an optional leading `+`/`-` sign:
  - sparse Stim `MPP`-style products: `"X0*X3*Z5*Y7"`, `"-Z0*Y1"`;
  - dense `"IXYZ…"` of length `n_qubits`, with `_` accepted as an alias for `I` so Stim `PauliString` text (`"+__X_Y_Z"`) parses directly.

  Malformed, out-of-range, or repeated-qubit observables produce a typed `ObservableParseError`.
- **`GeneralizedTableau::pauli_expectation(factors, negate)`**: returns `None` when any qubit in the observable's support has been lost; otherwise builds the word and delegates to the existing `expectation()` (#172). No state is disturbed.
- **`GeneralizedTableau::peek_observable_expectation(&str)`**: string front-end combining the two.
- **Python**: `GeneralizedTableau.peek_observable_expectation(observable)` returns an `ExpectationResult` — a frozen dataclass with a `LOST` sentinel, `is_lost`, and `__float__` (which raises on `LOST`). Complements the plain `expectation()`/`trace()` from #176 with sparse/signed syntax, input validation (`ValueError`), and loss-aware semantics.

## Relation to #172 / #176

This work originally carried its own multi-qubit decomposition path; it has been rebased to reuse `compute_decomposition_word` and `expectation()` from #172 instead (the only `data.rs` change left is making the `PhasedPauliWordNoHash` alias `pub(crate)`). What it adds on top is the observable-string grammar, lost-qubit handling, and the Python result type.

## Testing

- `cargo test -p ppvm-tableau` — all green (includes new parser tests in `observable.rs` and peek tests in `measure.rs`, alongside the existing #172 expectation tests); `cargo clippy` clean, `cargo fmt` clean.
- `pytest test/generalized_tableau/` — 170 passed (native module rebuilt), including `test_peek_expectation.py` and #176's `test_expectation.py` side by side.


rafaelha:
This is actually handled sufficiently well by the `expectation` API from https://github.com/QuEraComputing/ppvm/pull/172. The current API introduces inconsistent API surfaces. Closing it.

PR #214, opened 2026-08-21
chore(deps): updated `vihaco` dependency to `0.4.0`

robpatterson13:
### Summary
Migrates ppvm-vihaco and vihaco-circuit-isa from vihaco 0.1.1 to vihaco 0.4.0.

### Changes
- Migrates PPVM to vihaco 0.4.0 surface/runtime instruction APIs.
- Replaces the legacy parser pipeline with v0.4 parsing and resolution.
- Generates the circuit instruction set with `vihaco::component!`.
- Enforces one-to-one source-to-runtime instruction lowering.
- Updates .sst syntax, including:

```
  const.u64 0       → cpu::cpu.const u64, 0
  eq.u32            → cpu::cpu.eq u32
  circuit.h         → circuit::circuit.h
  ret               → cpu::cpu.ret 0
  @label:           → cpu::cpu.label @label
```

- Resolves branch targets and function calls to runtime addresses.
- Adds bytecode v2 serialization for function tables, labels, signatures, and main_function.
- Updates CLI, TUI, examples, embedded programs, and test fixtures.
- Adds and updates migration documentation in `docs/changelog/vihaco-0.4.md`.

### Verification
- cargo test --workspace
- cargo check -p ppvm-vihaco
- cargo test -p ppvm-vihaco
- cargo test -p vihaco-circuit-isa
- cargo test -p ppvm-cli
- cargo test -p ppvm-tui

### DIFF
--- a/crates/ppvm-cli/examples/ghz.sst
+++ b/crates/ppvm-cli/examples/ghz.sst
@@ -3,25 +3,25 @@ device circuit.n_qubits 3;
 // Prepare a 3-qubit GHZ state (|000> + |111>)/sqrt(2) and measure every qubit.
 // The three outcomes are perfectly correlated, so each shot reads 0 0 0 or 1 1 1.
 fn @main() {
-    const.u64 0
-    circuit.h
+    cpu::cpu.const u64, 0
+    circuit::circuit.h
 
-    const.u64 0
-    const.u64 1
-    circuit.cnot
+    cpu::cpu.const u64, 0
+    cpu::cpu.const u64, 1
+    circuit::circuit.cnot
 
-    const.u64 1
-    const.u64 2
-    circuit.cnot
+    cpu::cpu.const u64, 1
+    cpu::cpu.const u64, 2
+    circuit::circuit.cnot
 
-    const.u64 0
-    circuit.measure
+    cpu::cpu.const u64, 0
+    circuit::circuit.measure
 
-    const.u64 1
-    circuit.measure
+    cpu::cpu.const u64, 1
+    circuit::circuit.measure
 
-    const.u64 2
-    circuit.measure
+    cpu::cpu.const u64, 2
+    circuit::circuit.measure
 
-    ret
+    cpu::cpu.ret 0
 }
--- a/crates/ppvm-vihaco/Cargo.toml
+++ b/crates/ppvm-vihaco/Cargo.toml
@@ -19,9 +19,9 @@ num = "0.4.3"
 rayon = { version = "1.10", optional = true }
 ppvm-tableau = { version = "0.1.0", path = "../ppvm-tableau" }
 smallvec = "1.15.1"
-vihaco = "0.1.1"
-vihaco-cpu = "0.1.1"
-vihaco-parser = "0.1.1"
-vihaco-parser-core = "0.1.1"
+vihaco = "0.4.0"
+vihaco-cpu = "0.4.0"
+vihaco-parser = "0.4.0"
+vihaco-parser-derive = "0.4.0"
 vihaco-circuit-isa = { version = "0.1.0", path = "../vihaco-circuit-isa" }
 ppvm-pauli-sum = { version = "0.1.0", path = "../ppvm-pauli-sum" }
--- a/crates/ppvm-vihaco/src/bytecode.rs
+++ b/crates/ppvm-vihaco/src/bytecode.rs
@@ -233,11 +233,15 @@ fn encode_instruction(inst: &PPVMInstruction) -> eyre::Result<BytecodeInstructio
             CpuInstruction::Le(t) => BytecodeCpu::Le(encode_type(*t)),
             CpuInstruction::Ge(t) => BytecodeCpu::Ge(encode_type(*t)),
             CpuInstruction::Label(_) => {
-                return Err(eyre::eyre!("runtime labels are not serializable in PPVM bytecode"));
+                return Err(eyre::eyre!(
+                    "runtime labels are not serializable in PPVM bytecode"
+                ));
             }
         }),
         PPVMInstruction::Circuit(inst) => BytecodeInstruction::Circuit(match inst {
-            vihaco_circuit_isa::CircuitInstruction::TwoQubitPauliError => BytecodeCircuit::TwoQubitPauliError,
+            vihaco_circuit_isa::CircuitInstruction::TwoQubitPauliError => {
+                BytecodeCircuit::TwoQubitPauliError
+            }
             vihaco_circuit_isa::CircuitInstruction::Truncate => BytecodeCircuit::Truncate,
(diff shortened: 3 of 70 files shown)


crates/ppvm-cli/examples/ghz.sst
```diff
@@ -3,25 +3,25 @@ device circuit.n_qubits 3;
 // Prepare a 3-qubit GHZ state (|000> + |111>)/sqrt(2) and measure every qubit.
 // The three outcomes are perfectly correlated, so each shot reads 0 0 0 or 1 1 1.
 fn @main() {
-    const.u64 0
-    circuit.h
+    cpu::cpu.const u64, 0
```
david-pl, line 6: Could you explain why we need the module style markers here? It seems a bit redundant to have to prefix everything with `cpu::cpu.` and `circuit::circuit.`.

crates/ppvm-vihaco/src/bytecode.rs
```diff
@@ -12,12 +12,362 @@
...
+    Bool(bool),
+    I64(i64),
+    U32(u32),
+    U64(u64),
+    F64(f64),
+    FunctionRef(u32),
+    HeapRef(u32),
+}
+
+fn encode_type(value: Type) -> BytecodeType {
```
david-pl: Seems to me like these functions could be implemented via derive macros. It's a lot of boilerplate.

crates/ppvm-vihaco/src/composite.rs
```diff
@@ -27,13 +27,13 @@ use vihaco_circuit_isa::{CircuitEffect, CircuitInstruction, CircuitMessage};
 // existing `crate::composite::{…}` paths keep resolving.
 pub use crate::device_info::{BackendKind, PPVM_MAGIC, PPVMDeviceInfo};
 
-pub type Instruction = PPVMInstruction;
+type Instruction = ppvm::runtime::Instruction;
```
david-pl: This type is only used twice as far as I can tell. And it's an alias, so maybe remove that?

crates/ppvm-vihaco/Cargo.toml
```diff
@@ -19,9 +19,9 @@ num = "0.4.3"
...
 ppvm-tableau = { version = "0.1.0", path = "../ppvm-tableau" }
 smallvec = "1.15.1"
-vihaco = "0.1.1"
-vihaco-cpu = "0.1.1"
-vihaco-parser = "0.1.1"
-vihaco-parser-core = "0.1.1"
+vihaco = "0.4.0"
+vihaco-cpu = "0.4.0"
+vihaco-parser = "0.4.0"
+vihaco-parser-derive = "0.4.0"
```
david-pl, line 25: `cargo machete` says these can go, but I'm not sure, I think that was a false positive before.

docs/changelog/vihaco-0.4.md
```diff
@@ -0,0 +1,369 @@
...
+PPVM pushes operands with CPU constants, then executes one circuit instruction
+that consumes the required values from the CPU stack.
+
+| Old PPVM spelling | v0.4 spelling | Values consumed from stack |
+| --- | --- | --- |
+| `circuit.x` | `circuit::circuit.x` | qubit |
+| `circuit.y` | `circuit::circuit.y` | qubit |
+| `circuit.z` | `circuit::circuit.z` | qubit |
+| `circuit.h` | `circuit::circuit.h` | qubit |
+| `circuit.s` / `circuit.s_adj` | `circuit::circuit.s` / `circuit::circuit.s_adj` | qubit |
```
david-pl, line 252: Not sure if this is on purpose, but this is not consistent with the implementation. The new instructions are not snake cased (e.g. `s_adj` is now `sadj`). If we want to keep this md file, then we should fix it to avoid agent drift in the future.

david-pl (approved):
@robpatterson13 overall looks good, just some cosmetic comments. Agent review found two issues (which probably weren't introduced here, but still):

1. Symbolic function references such as `@helper` fail because FunctionRef operands are parsed only as numeric IDs.
2. main_function is stored as a code address instead of a function ID, and execution always starts at PC 0 rather than at `@main`.

Here are some failing tests (run as part of tests in `ppvm-vihaco/src/lib.rs`):

```rust
#[test]
fn symbolic_function_ref_calls_named_function() {
    let src = "device circuit.n_qubits 1;\n\
                fn @main() {\n\
                    cpu::cpu.const fn_ref, @helper\n\
                    cpu::cpu.call_indirect\n\
                    cpu::cpu.ret 0\n\
                }\n\
                fn @helper() {\n\
                    cpu::cpu.const u64, 0\n\
                    circuit::circuit.measure\n\
                    cpu::cpu.ret 0\n\
                }\n";

    let machine = run_program(src).expect("symbolic function reference should resolve");

    assert_eq!(machine.measurement_record().len(), 1);
}

#[test]
fn execution_starts_at_main_when_helper_is_declared_first() {
    let src = "device circuit.n_qubits 1;\n\
                fn @helper() {\n\
                    cpu::cpu.const u64, 0\n\
                    circuit::circuit.x\n\
                    cpu::cpu.ret 0\n\
                }\n\
                fn @main() {\n\
                    cpu::cpu.const u64, 0\n\
                    circuit::circuit.measure\n\
                    cpu::cpu.ret 0\n\
                }\n";

    let module = compile_program(src).expect("program should compile");
    let mut machine = PPVM::default();
    machine.load(&module).expect("module should load");
    machine.run().expect("program should run");

    assert_eq!(
        (module.main_function, machine.measurement_record().len()),
        (Some(1), 1),
        "main_function should be a function-table ID and execution should begin at @main"
    );
}
```

Also, CI is blocked because the pre-commit hooks fail in an unrelated crate (not sure why). Would you mind fixing that so CI actually runs on the PR?

I'll approve the changes, feel free to merge once the above things are cleared up.

crates/ppvm-cli/examples/ghz.sst
kaihsin: this is to support with nested composite component, say if I have a VM that contain two ppvm instances (or two cpus for example)

crates/ppvm-vihaco/Cargo.toml
robpatterson13: Updated the `cargo-machete` metadata

docs/changelog/vihaco-0.4.md
robpatterson13: Added the original spellings to the pattern parser

crates/ppvm-vihaco/src/bytecode.rs
robpatterson13: vihaco is going to handle file formats and bytecode encoding/decoding in the future, so most of this file will be handled by vihac-level features; for now I put small macros in the file to generate those encoding/decoding functions

PR #219, opened 2026-09-10
PPVM-traits-2 Patch

JonhasA:
A patch to move the changes to ppvm-trait from roger's PR https://github.com/QuEraComputing/ppvm/pull/204

Additionally, the following changes were made:
1. Added arithmetic module for defining traits to numeric types
2. added gates module for defining traits for Clifford/ channel / measure/ rotation ops
3. container module now has additional files related to storage or engine configuration
4. Moved PREFER_MOVED_RKEY outside of coefficient and into storage.rs in containers/. Engine will configure the key rather than having coefficient module be responsible for it.
5. Added a LossState trait to be associated with word sites rather than having it in PauliBits.
6. Isolated functionality related to fermionic factors into fermion_factor.rs
7. Pauli type is used explicitly for rotation operations
8. Moves container tests beside their implementations and adds regression coverage for batch hash invalidation and Pauli-channel factors.
9. Moves pauli_error_factors into the optional PauliErrorFactors channel capability, preserving the generic default and numeric specializations.

### DIFF
--- a/crates/ppvm-traits-2/src/algebra.rs
+++ b/crates/ppvm-traits-2/src/algebra.rs
@@ -1,0 +1,277 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Algebra capabilities that lift `C[K]` from a module to a (twisted) algebra:
+//! the key product [`KeyProduct`] and the two coefficient-ring capabilities it
+//! and the sesquilinear pairing need, [`ImaginaryUnit`] and [`Conjugate`], plus
+//! the [`Phase`] the Pauli key product emits.
+//!
+//! Design: `traits-2-configuration-and-hashing.md` §"The map is a graded
+//! algebra over `C[K]`" (L4 and its coefficient capabilities).
+
+use crate::arithmetic::Coefficient;
+
+/// A fourth root of unity `iᵏ`, `k ∈ ℤ/4`, i.e. an element of `{1, i, −1, −i}`.
+///
+/// The Pauli key product is not closed on keys: `v·w = iᵏ (v⊕w)`, so
+/// [`KeyProduct::key_mul`] returns the residual `Phase` for the coefficient to
+/// absorb. `iᵏ` already spans `{1, i, −1, −i}`, so no separate `±` is carried.
+///
+/// Design: §"The map is a graded algebra over `C[K]`" (`KeyProduct`). The
+/// exponent `k` is the packed `phaseExp` of `lean/PPVM/Pauli/Phase.lean`
+/// (`phaseExp_eq_ref`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Phase {
+    /// `i⁰ = +1`.
+    Pos1,
+    /// `i¹ = +i`.
+    PosI,
+    /// `i² = −1`.
+    Neg1,
+    /// `i³ = −i`.
+    NegI,
+}
+
+impl Phase {
+    /// The exponent `k ∈ {0, 1, 2, 3}` such that this phase equals `iᵏ`.
+    #[inline]
+    pub fn exponent(self) -> u8 {
+        match self {
+            Phase::Pos1 => 0,
+            Phase::PosI => 1,
+            Phase::Neg1 => 2,
+            Phase::NegI => 3,
+        }
+    }
+
+    /// The phase `iᵏ` for exponent `k` (taken mod 4).
+    #[inline]
+    pub fn from_exponent(k: u8) -> Self {
+        match k & 3 {
+            0 => Phase::Pos1,
+            1 => Phase::PosI,
+            2 => Phase::Neg1,
+            _ => Phase::NegI,
+        }
+    }
+
+    /// The group identity `i⁰ = +1` of the `ℤ/4` phase group.
+    ///
+    /// The neutral element for [`compose`](Self::compose): `p.compose(one()) ==
+    /// p` for every phase `p`.
+    #[inline]
+    pub fn one() -> Self {
+        Phase::Pos1
+    }
+
+    /// The `ℤ/4` group product `iᵃ · iᵇ = i^{a+b}` (exponents added mod 4).
+    ///
+    /// This is the first-class group operation on phases: a `KeyProduct` chain
+    /// can accumulate the residual phases each [`KeyProduct::key_mul`] emits with
+    /// `compose` (or the equivalent [`Mul`](std::ops::Mul) impl) without a coefficient in hand,
+    /// deferring the [`apply`](Self::apply) fold onto the coefficient to the end.
+    /// The group is abelian, so `a.compose(b) == b.compose(a)`.
+    #[inline]
+    pub fn compose(self, other: Self) -> Self {
+        Phase::from_exponent(self.exponent() + other.exponent())
+    }
+
+    /// The group inverse `i^{-k} = i^{4-k}`, i.e. the phase `q` with
+    /// `self.compose(q) == Phase::one()`.
+    #[inline]
+    pub fn inverse(self) -> Self {
+        // 4 - k is exact for k ∈ {0,1,2,3}; the mod-4 reduction in
+        // `from_exponent` sends k = 0 back to 0.
+        Phase::from_exponent(4 - self.exponent())
+    }
+
+    /// Fold this phase onto a coefficient: return `iᵏ · c`.
+    ///
+    /// This is the `iPow` fold of `lean/PPVM/Algebra/Twisted.lean` — the phase a
+    /// [`KeyProduct::key_mul`] emits is absorbed by the coefficient here. Needs a
+    /// primitive fourth root of unity, hence the [`ImaginaryUnit`] bound.
+    ///
+    /// # Behaviour parity
+    ///
+    /// The `±i` arms go through [`ImaginaryUnit::mul_i`], **not** through a bare
+    /// `c * imaginary_unit()`. On `Complex<f64>` the two are *not* the same
+    /// function: the ring multiply computes `(re·0 − im·1, re·1 + im·0)`, and
+    /// `inf·0`/`NaN·0` are `NaN`, so `(inf + 0i)·i` is `NaN + inf·i` while the
+    /// old `ppvm_traits::ComplexCoefficient::mul_phase` — which swapped the
+    /// components by hand — gave `−0 + inf·i`. `mul_i` restores the old
+    /// component swap (and with it the sign of zero, visible through `Display`
+    /// and serialization). See `phase_apply_matches_old_mul_phase_encoding`.
+    ///
+    /// The whole fold is delegated to [`ImaginaryUnit::mul_i_pow`], which is an
+    /// **override point**: a ring whose values carry `iᵏ` symbolically (the
+    /// symbolic `Term` of `ppvm-sym-2`) folds `iᵏ` into its own representation
+    /// rather than through the `±1`/`mul_i` arms, which is what old's
+    /// `ComplexCoefficient::mul_phase` did. The default body *is* those arms, so
+    /// nothing changes for `f64`/`Complex<f64>`/`GaussianInt`.
+    #[inline]
+    pub fn apply<C: ImaginaryUnit>(self, c: &C) -> C {
+        c.mul_i_pow(self.exponent())
+    }
+}
+
+/// `iᵃ · iᵇ = i^{a+b}` — [`compose`](Phase::compose) as the `*` operator, so a
+/// residual-phase accumulator reads `acc *= phase` / `acc = a * b`.
+impl core::ops::Mul for Phase {
+    type Output = Phase;
+
+    #[inline]
+    fn mul(self, rhs: Phase) -> Phase {
+        self.compose(rhs)
+    }
+}
+
+impl core::ops::MulAssign for Phase {
+    #[inline]
+    fn mul_assign(&mut self, rhs: Phase) {
+        *self = self.compose(rhs);
+    }
+}
+
+/// A key whose set carries a product — the (projective) group structure that
+/// lifts `C[K]` from a module to an algebra.
+///
+/// The keys form a group only *up to phase*: the product is **not closed on
+/// keys**, it emits an `iᵏ`, which is why `key_mul` returns `(Self, Phase)` and
+/// why `C[PauliWord]` is a **2-cocycle-twisted** group algebra.
+///
+/// # Laws
+///
+/// Write `key_mul(u, v) = (u · v, i^{β(u,v)})`. Every impl must satisfy:
+///
+/// * the key product is associative: `(u · v) · w == u · (v · w)`;
+/// * the phase exponent is a **2-cocycle**:
+///   `β(u,v) + β(u·v, w) == β(v,w) + β(u, v·w)` in `ℤ/4`.
+///
+/// Under exactly those two hypotheses the twisted product on `C × K` is
+/// associative for any commutative coefficient ring `C` with `i⁴ = 1` — proved
+/// key-agnostically in `lean/PPVM/Algebra/Twisted.lean` (`gtmul_assoc`, over an
+/// abstract `kmul` and `IsCocycle`), so the obligation is stated once and every
+/// key discharges it. `PauliWord` does so via `Bool.xor_assoc` and
+/// `lean/PPVM/Pauli/Phase.lean` (`phaseExp_cocycle`), recovered as the instance
+/// in `phaseExp_isCocycle` / `tmul_assoc_of_gtmul` (with `tmul_assoc` the
+/// concrete Pauli statement). A future ordered fermionic-word key must discharge
+/// the same two hypotheses; it does **not** inherit associativity from the Pauli
+/// proof.
+///
+/// Design: §"The map is a graded algebra over `C[K]`" (`KeyProduct`).
+pub trait KeyProduct: Eq + Clone {
+    /// Product of two keys, with the phase it produces (folded onto the coeff).
+    fn key_mul(&self, other: &Self) -> (Self, Phase);
+}
+
+/// The phase capability L4 needs, over a **commutative** coefficient ring: a
+/// distinguished primitive fourth root of unity `i`.
+///
+/// Impls must satisfy `Self::imaginary_unit() * Self::imaginary_unit() ==
+/// -Self::one()` (hence `i⁴ = 1`). This is strictly weaker than requiring
+/// `Complex<f64>`: `GaussianInt` (`ℤ[i]`), `Complex<Rational>`, and cyclotomic
+/// integers all satisfy it, so L4 does not foreclose exact Pauli multiplication.
+///
+/// Design: §"The map is a graded algebra over `C[K]`" (`ImaginaryUnit`). Law
+/// machine-checked in `lean/PPVM/Pauli/Matrix.lean` (`iU_sq`: `iU * iU = -1`),
+/// and the twisted product is associative over any commutative ring with
+/// `i⁴ = 1` in `lean/PPVM/Algebra/Twisted.lean` (`tmul_assoc`).
+pub trait ImaginaryUnit: Coefficient + num::One {
+    /// The imaginary unit `i`; impls must satisfy
+    /// `Self::imaginary_unit() * Self::imaginary_unit() == -Self::one()`.
+    fn imaginary_unit() -> Self;
+
+    /// Multiply by `i`. Semantically `self * imaginary_unit()`, which is the
+    /// default body — but it is an **override point**, because on a
+    /// floating-point ring the generic product is not extensionally equal to the
+    /// rotation it denotes.
+    ///
+    /// On `Complex<f64>`, `c * i` expands to
+    /// `(re·0 − im·1, re·1 + im·0)`; `inf·0` and `NaN·0` are `NaN`, so a
+    /// non-finite component contaminates *both* output components, and `re·0`
+    /// also loses the sign of zero. Multiplication by `i` is really the
+    /// component swap `(re, im) ↦ (−im, re)`, which is total and exact — this is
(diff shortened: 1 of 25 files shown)


crates/ppvm-traits-2/src/arithmetic/coefficient.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// Multiply this coefficient in place by `sign ∈ {-1, +1}`.
+    #[inline]
+    fn mul_sign_assign(&mut self, sign: i8) {
+        *self = self.mul_sign(sign)
+    }
+
+    ///  Accumulates a borrowed coefficient.
+    #[inline]
+    fn add_assign_ref(&mut self, rhs: &Self) {
```
david-pl, line 33: This isn't used anywhere (but it should be). Why don't we just require

```rust
for<'a> AddAssign<&'a Self>  + for<'a> MulAssign<&'a Self>
```

on the `Coefficient` trait? These should be fulfilled and then you can just do 

```rust
c1 += &c2
```

without cloning.

crates/ppvm-traits-2/src/arithmetic/halvable.rs
```diff
@@ -0,0 +1,26 @@
...
+/// apply the `(I ± Z)/2` projectors.
+pub trait Halvable: Coefficient {
+    /// Divide by two. Impls must be exact: `x.half() + x.half() == x`.
+    fn half(&self) -> Self;
+}
+
+impl Halvable for f64 {
+    #[inline]
+    fn half(&self) -> Self {
+        *self / 2.0
```
david-pl, line 17: I'm surprised that this holds up in the exact `x.half() + x.half() == x` condition.

crates/ppvm-traits-2/src/containers/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+    fn toggled_bits(&self, row: usize, qubit: usize, toggle_x: bool, toggle_z: bool) -> Self::Key
+    where
+        Self::Key: PauliBits,
+    {
+        self.get(row).toggled_bits(qubit, toggle_x, toggle_z)
+    }
+
+    /// Materialize one row while toggling bits at two sites.
+    #[inline]
+    #[allow(clippy::too_many_arguments)]
```
david-pl: Never a fan of seeing clippy exceptions.

crates/ppvm-traits-2/src/containers/batch.rs
```diff
@@ -0,0 +1,460 @@
...
+pub trait Columnar: Indexable {
+    /// The concrete structure-of-arrays column for this key type.
+    type Column: KeyColumn<Key = Self>;
+}
+
+/// A structure-of-arrays column of keys, owned by the concrete key type (only it
+/// knows its planes). Operates plane by plane, never scalar on the hot path.
+///
+/// Design: §"The batch contract".
+pub trait KeyColumn: Default + Clone {
```
david-pl, line 17: Overall, this trait seems a bit complex. Can we split it maybe?

crates/ppvm-traits-2/src/containers/graded.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// `(&keys[i], &coeffs[i])`, which needs no materialized pair.
+    ///
+    /// The default body goes through `iter`, so this is a capability an existing
+    /// backend may ignore; every backend whose keys and coefficients are actually
+    /// *stored* should override it, and the shipped ones do.
+    ///
+    /// It is a scan, not an iterator, deliberately: returning
+    /// `impl Iterator<Item = (&K, &C)>` would force every backend to have a
+    /// lending iterator, and a callback composes with the `Retain`/`Policy`
+    /// closures already used on this side of the API.
```
david-pl: Can you tell claude to go a bit easy on the comments? 😅 Not sure if this is copy & paste from #204 but comments are quite excessive throughout.

crates/ppvm-traits-2/src/containers/storage.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/// Implemented by storage backends, independently of coefficient arithmetic.
+/// Both traversals must preserve the same keys and coefficient values.
+pub trait RekeyStrategy {
```
david-pl: I get why this is here, but it's still kind of odd to add an orphaned trait. I don't think it's used anywhere in this PR.

crates/ppvm-traits-2/src/gates/channel.rs
```diff
@@ -0,0 +1,257 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::arithmetic::Coefficient;
+
+/// Coefficients that can calculate single-qubit Pauli-channel factors.
+///
+/// This optional channel capability keeps noise formulas out of scalar
+/// arithmetic. Implementers may use the generic default or specialize it.
+pub trait PauliErrorFactors: Coefficient + num::One {
```
david-pl, line 8: This seems a bit overly specific. I think it's still reasonable to require bounds that you can do simple math on coefficients (e.g. having `One + Add + Sub + Mul`). Also, it seems this trait is only for `PauliError`. How are we doing this for e.g. loss?

crates/ppvm-traits-2/src/pauli.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// Multiply generator `src` into `dst` (uses the Aaronson–Gottesman
+    /// `g`-rule).
+    fn row_multiply(&mut self, src: usize, dst: usize);
+
+    /// Restore canonical form after elimination.
+    fn canonicalize(&mut self);
+}
+
+pub trait BlanketClifford {}
```
david-pl: Maybe add some docs as to why this is needed.

crates/ppvm-traits-2/src/gates/rot.rs
```diff
@@ -0,0 +1,248 @@
...
+    fn ry(&mut self, qubit: usize, theta: A) {
+        self.rotate_1(Pauli::Y, qubit, theta)
+    }
+    /// Rotate about `Z` on `qubit` by `theta`.
+    fn rz(&mut self, qubit: usize, theta: A) {
+        self.rotate_1(Pauli::Z, qubit, theta)
+    }
+
+    /// Explicit batched `RX(θ)`.
+    fn rx_many(&mut self, targets: &[usize], theta: A)
```
david-pl, line 34: I know this is "historic" - as in, it's currently the same on the main branch - but: having the rotation `*_many` methods live on the same trait whereas we have an explicit `CliffordBatch` trait is kind of inconsistent.

crates/ppvm-traits-2/src/gates/clifford.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+    /// stim alias for [`cz`](Clifford::cz).
+    fn zcz(&mut self, qubit0: usize, qubit1: usize) {
+        self.cz(qubit0, qubit1)
+    }
+}
+
+/// Additional Clifford gates beyond the minimal set: `S†`, `√X`, `√X†`, `√Y`,
+/// `√Y†`, and `CY`.
+pub trait CliffordExtensions: Clifford {
```
david-pl: While we're refactoring stuff: I would suggest to merge `CliffordExtensions` into `Clifford`. I don't think there's any struct implementing one but not the other. I just added that as a quick fix at some point while writing the tableau implementation so I wouldn't have to add those methods to `PauliSum`.

crates/ppvm-traits-2/src/algebra.rs
```diff
@@ -0,0 +1,277 @@
...
+    Neg1,
+    /// `i³ = −i`.
+    NegI,
+}
+
+impl Phase {
+    /// The exponent `k ∈ {0, 1, 2, 3}` such that this phase equals `iᵏ`.
+    #[inline]
+    pub fn exponent(self) -> u8 {
+        match self {
```
david-pl: Compiles to the same thing, but you could just to `self as u8` here.

crates/ppvm-traits-2/src/fermion_factor.rs
```diff
@@ -0,0 +1,22 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/// The action a fermionic factor performs on its mode.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum FermionAction {
```
david-pl: Is this used anywhere?

crates/ppvm-traits-2/src/fermion_factor.rs
```diff
@@ -0,0 +1,22 @@
...
+    Create,
+    /// An annihilation operator `a`.
+    Annihilate,
+}
+
+/// One factor of an ordered fermionic product — the alphabet of a future
+/// ordered fermionic word (`Word<Site = FermionSite>`), whose index denotes
+/// factor order while the site carries the physical mode.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct FermionSite {
```
david-pl: Same here: is that used?

crates/ppvm-traits-2/src/loss.rs
```diff
@@ -0,0 +1,31 @@
+// SPDX-FileCopyrightText: 2026 The PPVM Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Loss state associated with word sites
+pub trait LossState {
```
david-pl, line 5: For reference: #195 adds leakage support to `GeneralizedTableau`. It currently uses an enum to represent that state. Should we decide to support leakage in PauliSum, this might be the point where it enters.

crates/ppvm-traits-2/src/word.rs
```diff
@@ -0,0 +1,214 @@
...
+        out
+    }
+
+    /// Consume a word and toggle X/Z bits at two sites in place.
+    ///
+    /// Re-keying kernels already own their key, so this avoids the structural
+    /// copy required by [`toggled_bits2`](PauliBits::toggled_bits2). Packed words
+    /// may override it to defer derived-hash refresh until all writes complete.
+    #[inline]
+    #[allow(clippy::too_many_arguments)]
```
david-pl: Again the clippy exception here. I think it might make sense to split up this trait or at least this function.

david-pl (changes requested):
Overall a solid first breakdown of the huge PR, but I think we should make some changes here.

Also, on a more general note: while I get that proving things with lean is beneficial, I think this introduces unneeded complexity at points. For example, to fulfill criteria so this is provably a ring we introduce some traits that are sometimes a bit odd, e.g. `Halvable`.
cc @Roger-luo

crates/ppvm-traits-2/src/containers/graded.rs
JonhasA: these are a copy and paste from https://github.com/QuEraComputing/ppvm/pull/204 lol I'll go ahead and clean up some of the comments

crates/ppvm-traits-2/src/fermion_factor.rs
JonhasA: Within the crate itself, the fermion_factor module is not used anywhere within it. However, I wasn't sure if we want to keep it or not.

crates/ppvm-traits-2/src/fermion_factor.rs
JonhasA: This was removed. It remained back from 204, but was removed here. If required, we could always add it back.

crates/ppvm-traits-2/src/pauli.rs
JonhasA: So instead of keeping it, I just removed it. The reason it was there initially was for types that are bounded by  SymplecticColumns + PhaseTrack can just use the empty Clifford trait and inherit the methods shown a couple lines below. However, depending on the consumer or some other backend, such implementations may not be appropriate for that consumer.

Therefore, 
-impl<T: SymplecticColumns + PhaseTrack + BlanketClifford> Clifford for T { ... }
was removed and the BlanketClifford trait was removed as well to force the type consumer to implement their own methods.

Roger-luo:
> Also, on a more general note: while I get that proving things with lean is beneficial, I think this introduces unneeded complexity at points. For example, to fulfill criteria so this is provably a ring we introduce some traits that are sometimes a bit odd, e.g. Halvable.


For reference, One use case is symbolic expressions. The minimum requirement of a symbolic expression is that it can be divided by half, and it doesn't matter what number that half comes from.

crates/ppvm-traits-2/src/containers/batch.rs
Roger-luo: When split into a package or a separate module?

crates/ppvm-traits-2/src/gates/channel.rs
Roger-luo: Well, when I was creating this trait, I must admit I didn't find a good way to generalize everything. 

It probably makes sense to just assume all the error channel parameters are under the same trait, but you still kind of need a different set of traits for each error channel family so they satisfy certain constraints.

Roger-luo: The loss channel coefficient doesn't really require this constraint. If you look at the loss channel trait defined below, it's just taking a regular coefficient trait.

crates/ppvm-traits-2/src/gates/channel.rs
```diff
@@ -0,0 +1,248 @@
...
+/// Correlated two-qubit loss channel.
+pub trait CorrelatedLossChannel<C: Coefficient> {
+    /// Apply correlated loss: `p[0]` loses both qubits when both are present;
+    /// `p[1]` loses either one when both are present; `p[2]` loses the remaining
+    /// qubit when the other was already lost.
+    fn correlated_loss_channel<R: rand::Rng + ?Sized>(
+        &mut self,
+        qubit0: usize,
+        qubit1: usize,
+        p: [C; 3],
```
Roger-luo, line 200: This might need similar treatment, like the pauli channrl error coefficient factors, too, because we assume the three parameters also satisfy the normalization condition, right?

crates/ppvm-traits-2/src/gates/clifford.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// Apply `CY` to one `(control, target)` pair.
+    fn cy(&mut self, control: usize, target: usize);
+    /// stim alias for [`cy`](Clifford::cy).
+    fn zcy(&mut self, control: usize, target: usize) {
+        self.cy(control, target)
+    }
+}
+
+/// Batched Clifford gates: apply the same gate to many qubits in one call.
+pub trait CliffordBatch: Clifford {
```
Roger-luo, line 53: I was thinking if there is a way to generalize this based on the Clifford trait, but I didn't really find a good way to define a broadcast trait in general, so I ended up doing this.

crates/ppvm-traits-2/src/gates/channel.rs
david-pl: > Well, when I was creating this trait, I must admit I didn't find a good way to generalize everything.

I would argue that it's worth the simplicity of having all noise parameters live under a single trait at the cost of being a bit too restrictive for some of the noise channels (like the uncorrelated loss). After all, when would you ever expect running something with a coefficient that supports just one, but not the other?

crates/ppvm-traits-2/src/arithmetic/coefficient.rs
```diff
@@ -0,0 +1,82 @@
...
```
david-pl: My point before was to require that trait and then remove this method. It's much more natural to just write `*self += rhs` and it's also done for the most part (which is why this method was only used once). So, just remove it?

crates/ppvm-traits-2/src/containers/batch.rs
david-pl: Just split it up into 2 or 3 separate traits. Could be a module, I don't think we need another crate.

crates/ppvm-traits-2/src/gates/clifford.rs
david-pl: Related question: do we have any consumers of this trait that don't run a custom impl? Do we really need the fallback loops?

crates/ppvm-traits-2/src/algebra.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /// The phase `iᵏ` for exponent `k` (taken mod 4).
+    #[inline]
+    pub fn from_exponent(k: u8) -> Self {
+        match k & 3 {
+            0 => Phase::Pos1,
+            1 => Phase::PosI,
+            2 => Phase::Neg1,
+            _ => Phase::NegI,
```
david-pl: This could silently introduce false results if there's a bug somewhere such that `k & 3 > 3`.

david-pl (commented):
Some minor stuff I still found, and some responses to @Roger-luo here. But it's almost there.

crates/ppvm-traits-2/src/containers/batch.rs
Roger-luo: I see. Yeah, this sounds reasonable to me. The key column might be a little bit overcomplicated.

crates/ppvm-traits-2/src/gates/clifford.rs
Roger-luo: I think technically you could have a batch version of the gate implementation, with the consideration of SIMD and parallel operation. For example, if you are applying a batch with a static, known size ahead of time, say it is an X gate, you could come up with the flipping rules ahead of time that modify a batch of bits instead of a single bit in advance. 

In principle, that would be faster because then you can utilize SIMD operations and parallel access to your containers.

crates/ppvm-traits-2/src/algebra.rs
JonhasA: in https://github.com/QuEraComputing/ppvm/pull/204, `Sum<S, P>` overrides only `cnot_many` and inherits the loops for `x_many`, `y_many`, and the other _many operators.

crates/ppvm-traits-2/src/containers/hash_join.rs
```diff
@@ -0,0 +1,203 @@
...
+    K: Indexable,
+    C: Coefficient,
+{
+    /// Build side of the hash join: probe each produced term, accumulate its
+    /// coefficient onto a matching key, insert on a miss.
+    #[inline]
+    fn accumulate_batch(&mut self, terms: &TermBatch<K, C>) {
+        for (k, c) in terms.iter() {
+            self.entry(k.clone())
+                .and_modify(|e| *e += c.clone())
```
david-pl: This no longer needs to `c.clone()` now due to the `AddAssign` bound we added.

crates/ppvm-traits-2/src/containers/coordinate_list.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn len(&self) -> usize {
+        self.as_slice().len()
+    }
+
+    #[inline]
+    fn get(&self, key: &K) -> Option<C> {
+        self.as_slice()
+            .iter()
+            .find(|(k, _)| k == key)
+            .map(|(_, c)| c.clone())
```
david-pl: `c.clone()` is no longer needed because of the `AddAssign` bound.

crates/ppvm-traits-2/src/arithmetic/halvable.rs
david-pl: Actually, there is an edge case here: this doesn't hold for `f64::NAN`, which is a perfectly valid `f64`, but `x.half() == x` if `x` is `NAN`.

crates/ppvm-traits-2/src/gates/channel.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+/// A unital single-qubit Pauli error channel `P ↦ λ_P·P`.
+pub trait PauliError<C: Coefficient> {
+    /// Apply a single-qubit Pauli channel with `X`, `Y`, `Z` probabilities.
+    fn pauli_error<R: rand::Rng + ?Sized>(
+        &mut self,
+        qubit: usize,
+        probabilities: [C; 3],
+        rng: &mut R,
```
david-pl: Actually, `PauliSum` is deterministic in noise channels. Why are we passing in an `rng` here? `PauliSum` doesn't have one. @Roger-luo ?

crates/ppvm-traits-2/src/gates/channel.rs
```diff
@@ -0,0 +1,248 @@
...
+/// (see [`LossState`](crate::loss::LossState)).
+pub trait LossChannel<C: Coefficient> {
+    /// Apply a loss channel to `qubit` with loss probability `p`.
+    fn loss_channel<R: rand::Rng + ?Sized>(&mut self, qubit: usize, p: C, rng: &mut R);
+}
+
+/// Correlated two-qubit loss channel.
+pub trait CorrelatedLossChannel<C: Coefficient> {
+    /// Apply correlated loss: `p[0]` loses both qubits when both are present;
+    /// `p[1]` loses either one when both are present; `p[2]` loses the remaining
```
david-pl: There was a recent fix to the correlated loss convention here: https://github.com/QuEraComputing/ppvm/pull/205

This docstring seems to have regressed now, probably because #204 was based off the old state.

david-pl (commented):
The left-over `clones` need to go. And I just realized that there's no rng on `PauliSum`. Why was it added to the trait @Roger-luo ?

Roger-luo:
good catch, this is a part I'm not sure - I moved `rng` out because it's an external state not necessarily (should) be managed by the PauliSum/tableau which is a quantum state, I think it's more natural to have it passed as a argument and let external/global managing it so it's more explicit who is sharing the RNG with what seed.

On the other hand, RNG is technically part of the entire execution state, so it could make sense to let the state object carrying it.

PR #221, opened 2026-09-15
Leakage bytecode instruction

david-pl:
After #195 gets merged, we need to rebase this branch with

```bash
git rebase --onto origin/main david/leakage david/leakage-bytecode
```

cc @kaihsin this PR introduces a breaking change in the vihaco-circuit-isa since it adds the enum variants for leakage.


Roger-luo (approved):
I think this PR is quite straightforward, but at some point we probably need to separate noise instructions from regular gates, because you don't actually need the noise for power estimation. 

If we're actually running things in simulation mode, you would expect something to automatically insert the noise instructions as part of the noise modelling.

